Issue #4, opened 2024-08-20
🔧 Add docker container support


FranklinWaller:
## 🔧 Chore

Add support for a docker

## Motivation

Allows data providers to easily run a container with the settings they want

kaynetik:
If I understood the intention behind this ticket, we want to enable data providers to initiate `config.json` to their own liking?

If that assumption is correct, then please verify that the following statements cover all possible statements @Thomasvdam .

## 1. Basic Structure
```json
{
  "routeGroup": "string",
  "routes": [
    {
      "path": "string",
      "upstreamUrl": "string",
      "method": "string" | ["string"],
      "headers": {
        "header-name": "string"
      },
      "jsonPath": "string",
      "forwardResponseHeaders": ["string"]
    }
  ]
}
```

## 2. Route Configuration

### 2.1 Path
- Simple path: `"/eth-usd"`
- Variable path: `"/:coinA/:coinB"`
- Wildcard path: `"/*"`

### 2.2 Upstream URL
- Simple URL: `"https://myapi.com/eth-usd"`
- URL with variables: `"https://myapi.com/{:coinA}-{:coinB}"`
- URL with wildcard: `"https://swapi.dev/api/{*}"`

### 2.3 Method
- Default (GET): Omit the "method" field
  - Or was this a wrong assumption? Should we require a method to be specified?
- Single method: `"method": "GET"`
- Multiple methods: `"method": ["GET", "HEAD"]`

### 2.4 Headers
```json
"headers": {
  "x-api-key": "string",
  "x-custom": "string"
  // Do we leave room for custom key-value header pairs?
}
```

### 2.5 JSON Path
```json
"jsonPath": "$.name"
```

### 2.6 Forward Response Headers
```json
"forwardResponseHeaders": ["content-type", "server"]
```

## 3. Variable Injection

### 3.1 Route Variables
- In path: `"/:varName"`
- In upstream URL: `"{:varName}"`
- In headers: `"x-custom": "{:varName}"`

### 3.2 Environment Variables
- In headers: `"x-api-key": "{$SECRET_API_KEY}"`
- Can be used in any string field

### 3.3 Wildcard
- In path: `"/*"`
- In upstream URL: `"{*}"`

## 4. Example
```json
{
  "routeGroup": "proxy",
  "routes": [
    {
      "path": "/:coinA/:coinB",
      "upstreamUrl": "https://myapi.com/{:coinA}-{:coinB}",
[...]

PR #22, opened 2024-10-04
feat: Dockerize DP 🐳

kaynetik:
TODOs:
- [x] Centralize all build efforts (probably `docker/build`)
  - [x] Enable makefile
- [x] Verify entry point script with valid configs
- [x] **Create GHA release pipeline**

## Commits

- **feat: GHA Build & Release :rocket:**
- feat: Introduce pre-commit hook
- **refactor: Improve entrypoint 🐳**
- feat: Established an entrypoint script for DP runner
- feat: Run DP with Docker

-----

## Motivation

(Write your motivation here)

## Explanation of Changes

(Write your explanation here)

## Testing

(Write your test plan here)

## Related PRs and Issues

- Closes: #4 
- Unblocks: [#6](https://github.com/sedaprotocol/iac-gcloud/issues/6)


mariocao (approved):
Looks good to me, maybe add some docker instructions in the readme, but we can open an issue for that and do it later. Your call.

.pre-commit-config.yaml
Thomasvdam, line 1: Unless we standardise on something as a team I would expect these files to be ignored. In the explorer we use https://commitlint.js.org/guides/local-setup.html and Husky to manage git hooks, but given that a more diverse group works in this repo we should probably discuss it first :P

.build/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+elif [ "$CONFIG_EXISTS" = true ] && [ "$PK_EXISTS" = false ]; then
+  # Only config.json is provided
+  echo "Running with config only"
+  run_bun_command start init
+  RUN_CMD="start run --config /app/config.json"
+else
+  # Neither config.json nor private key is provided
+  echo "Running with --disable-proof"
+  run_bun_command start init
+  RUN_CMD="start run --disable-proof"
```
Thomasvdam, line 49: This doesn't make sense: `--disable-proof` should only skip validating whether the incoming request has the correct proof, it doesn't deal with the private key or config at all.

.build/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+# Determine the command to run based on the presence of config.json and private key
+if [ "$CONFIG_EXISTS" = true ] && [ "$PK_EXISTS" = true ]; then
+  # Both config.json and private key are provided
+  echo "Running with config and private key"
+  RUN_CMD="start run --config /app/config.json --private-key-file /app/data-proxy-private-key.json"
+elif [ "$CONFIG_EXISTS" = true ] && [ "$PK_EXISTS" = false ]; then
+  # Only config.json is provided
+  echo "Running with config only"
+  run_bun_command start init
+  RUN_CMD="start run --config /app/config.json"
```
Thomasvdam, line 44: I think the binary takes care of these checks, is there a reason to also do it in the entrypoint?

.build/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Check if data-proxy-private-key.json exists
+if [ -f /app/data-proxy-private-key.json ]; then
+  echo "data-proxy-private-key.json detected"
+  PK_EXISTS=true
+elif [ -n "$SEDA_DATA_PROXY_PRIVATE_KEY" ]; then
+  # If private key file does not exist, check if the private key is provided via environment variable
+  echo "Private key provided via environment variable"
+  echo "$SEDA_DATA_PROXY_PRIVATE_KEY" >/app/data-proxy-private-key.json
+  PK_EXISTS=true
```
Thomasvdam, line 23: I don't think this will work, as the key file is a JSON file with both the public and private key. I'd imagine most people would only supply the private key as an environment variable.

.build/docker/docker-entrypoint.sh
Thomasvdam: The data-proxy can not run when there is no config or no private key (file). `--disable-proof` only disables the request verification, so it should be a completely separate option.

I think the scenarios should be covered by the binary itself where it will error if it is unable to find a config and/or private key (file).

If someone wants to run it in debug mode they should pass the `--disable-proof` flag manually.

.pre-commit-config.yaml
Thomasvdam: Probably in the engineering Slack channel, or opening a PR that introduces the tool.

The reason I commented on this file is that if we want to commit this file (and this workflow) we should also document it in the repo, otherwise it's just clutter.

.build/docker/docker-entrypoint.sh
Thomasvdam: If the env var is set the data-proxy uses that to construct the private/public key pair and won't even attempt to read the key file.

.build/docker/docker-entrypoint.sh
Thomasvdam: I'm not entirely sure what the best approach would be here... 🤔 

I'm partial to letting the binary take care of everything as this reduces the maintenance burden on our end since the container doesn't really need to know anything.

If I want to initialise I should do `docker run <IMAGE_NAME> init` and pass a volume so I can persist the config and generated key.
For registration I can then do `docker run <IMAGE_NAME> register <PAYOUT_ADDRESS> <AMOUNT>` and click the link from the output to complete the registration.
And to run it just do `docker run <IMAGE_NAME> run`.

I guess that get a little messy with compose though, don't know how to do that in an elegant fashion.

.build/docker/docker-entrypoint.sh
mariocao: yes, passing it manually makes sense, maybe we could have an container env variable like `FLAGS` or so..

PR #40, opened 2025-02-24
feat: Migration to GCP build 🐳

kaynetik:
## Commits

- **feat: Init Cloud Build for DP 🚀**
- **fix: Apply cloud logging (if available)**
- **refactor: Add step-by-step Docker auth**
- **chore: Cleanup CloudBuild for DataProxy**
- **chore: Cleanup comments**

## Motivation

This was a required step for the simplification of our startup-scripts model.

## Explanation of Changes

This commit removes the usage of GHA for generation of a release, and switches to CloudBuild from GCP, which in turn uses `gh` cli to create a public release. So the end user won't be affected at all, and our images will be stored in the Artifactory.

## Testing

Was tested on our `perf-testing` environment. Though some improvements will come in the followup PRs.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -100,7 +100,7 @@ steps:
 #
 # Step 6: Create GitHub Release
 #
-- name: 'node:18'
+- name: 'node:20'
   id: 'create-release'
   waitFor: ['create-manifest']
   secretEnv: ['GITHUB_TOKEN']


cloudbuild.yaml
```diff
@@ -0,0 +1,162 @@
...
+      docker manifest create \
+        ${_REGISTRY}/${_PROJECT}/${_REPO}/${_IMAGE_NAME}:${TAG_NAME} \
+        --amend ${_REGISTRY}/${_PROJECT}/${_REPO}/${_IMAGE_NAME}:${TAG_NAME}-amd64 \
+        --amend ${_REGISTRY}/${_PROJECT}/${_REPO}/${_IMAGE_NAME}:${TAG_NAME}-arm64
+      docker manifest push ${_REGISTRY}/${_PROJECT}/${_REPO}/${_IMAGE_NAME}:${TAG_NAME}
+
+#
+# Step 6: Create GitHub Release
+#
+- name: 'node:18'
```
Thomasvdam, line 103: Should we not use a version that’s still in the maintenance window?

mariocao: let's replace it to `node:20`

mariocao (approved):
Change node version to 20 and then you can merge it :)

PR #57, opened 2025-06-12
Conditionally retry proof verification 

Thomasvdam:
## Motivation

We've observed some problems where it seems that the RPC which the
    overlay uses to query eligibility is ahead of the the RPC that the data
    proxy uses to query eligibility. We've added extra log statements to
    help debug this as well as retry logic when we have reason to believe
    the RPC is out of sync.

## Explanation of Changes

Only retry when we have reason to believe that the RPC is out of sync

## Testing

Will add unit tests for the retry behaviour

## Related PRs and Issues

N.A.

### DIFF
--- a/workspace/data-proxy-sdk/src/constants.ts
+++ b/workspace/data-proxy-sdk/src/constants.ts
@@ -1,4 +1,5 @@
 export const PROOF_HEADER_KEY = "x-seda-proof";
+export const HEIGHT_HEADER_KEY = "x-seda-blockheight";
 export const SIGNATURE_HEADER_KEY = "x-seda-signature";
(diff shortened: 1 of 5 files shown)


workspace/data-proxy-sdk/src/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const PROOF_HEADER_KEY = "x-seda-proof";
+export const HEIGHT_HEADER_KEY = "x-seda-height";
```
mariocao, line 2: Just being picky...
```suggestion
export const HEIGHT_HEADER_KEY = "x-seda-blockheight";
```

Issue #26, opened 2024-10-21
✨ Add CLI command to show public key


Thomasvdam:
## ✨ Feature

Add CLI command to show public key.

## Motivation

Make it a little easier to see which identity is being used for the data proxy.

## Implementation

Display the public key for the data proxy if one is registered. Maybe good to also print a link to the explorer page for that proxy?

hacheigriega:
closed by #61

PR #61, opened 2025-06-18
Command for validating node config

### LINKED ISSUE #50: ✨ Add a CLI command to verify config
## ✨ Feature

As a user it'd be useful to have a command that verifies the config/setup is correct without immediately running the node. It would be great if it could also print out the config it's going to use while masking any sensitive values.

## Motivation

Make it easier for users to set up the node and preview config changes before committing to them.

## Implementation

- [ ] Add new CLI command that prints the config the node is going to use
- [ ] Make sure to mask any sensitive values. There should be a difference between a present secret value and one that's missing.
- [ ] Add a flag that skips the printing and only validates that it is valid

hacheigriega:
## Explanation of Changes

```
❯  bun run start validate
$ bun run check-ts && bun run ./workspace/data-proxy validate
$ bunx tsc --noEmit
2025-06-18 16:11:35.391 info: Using config: ./config.json
🌐 Network: testnet
🔐 Using public key: 02f2f92d36ad5e6613c1df210ba9a68272375926a605b716b4cd30d49af9910129
🎟️  Registration info: {
  "payoutAddress": "seda1lyfkj6xwecq3462qa683s78weexhp7d72ku3qf",
  "fee": {
    "denom": "aseda",
    "amount": "10000000"
  },
  "memo": "",
  "adminAddress": "seda1lyfkj6xwecq3462qa683s78weexhp7d72ku3qf"
}
⚙️  Config: {
  "routeGroup": "proxy",
  "routes": [
    {
      "baseURL": {
        "variant": "Nothing"
      },
      "path": "/:symbolA/:symbolB",
      "upstreamUrl": "https://api.binance.com/api/v3/ticker/price?symbol={:symbolA}{:symbolB}",
      "method": [
        "GET"
      ],
      "forwardResponseHeaders": {},
      "headers": {}
    },
    {
      "baseURL": {
        "variant": "Nothing"
      },
      "path": "mock/:target",
      "upstreamUrl": "http://1.1.1.1:8080/api/{:target}",
      "method": [
        "GET"
      ],
      "forwardResponseHeaders": {},
      "headers": {}
    }
  ],
  "baseURL": {
    "variant": "Nothing"
  },
  "statusEndpoints": {
    "root": "status"
  }
}
✅ SEDA Data Proxy configuration is valid
```

Supports `--silent` flag for silencing config printout and `--skip-registration-check` flag for skipping registration check.

## Related PRs and Issues

Closes #50


Thomasvdam (changes requested):
I think this needs a little more work as the warnings are not super clear right now:

```
2025-06-19 09:52:33.820 info: Using config: ./config.json
2025-06-19 09:52:33.822 warn: $.0.path has set :target but was not used in $.0.upstreamUrl or $.0.headers.
Please either remove the variable from the path or use it in the upstreamUrl or headers (through "{:target}").
🌐 Network: planet
🔐 Using public key: 03108a3eb57c53acd84fe99cb70da0cac32107a37df589b16e8473819dad9d10fa
🎟️  Registration info: {
  "payoutAddress": "seda1lyfkj6xwecq3462qa683s78weexhp7d72ku3qf",
  "fee": {
    "denom": "aseda",
    "amount": "10000000000000000"
  },
  "memo": "This is a sweet proxy 7",
  "adminAddress": "seda1lyfkj6xwecq3462qa683s78weexhp7d72ku3qf"
}
⚙️  Config: {
  "routeGroup": "proxy",
  "routes": [
    {
      "baseURL": {
        "variant": "Nothing"
      },
      "path": "mock/:target",
      "upstreamUrl": "http://3.8.215.180:8080/api/{target}",
      "method": [
        "GET"
      ],
      "forwardResponseHeaders": {},
      "headers": {}
    }
  ],
  "baseURL": {
    "variant": "Nothing"
  },
  "statusEndpoints": {
    "root": "status"
  }
}
✅ SEDA Data Proxy configuration is valid
```

I think the final status should at least indicate that there's a warning.

This is maybe also a good opportunity to address the TODO in `workspace/data-proxy/src/utils/replace-params.ts` and check the ENV var replacement during config parsing.

hacheigriega:
> I think the final status should at least indicate that there's a warning.
> 
> This is maybe also a good opportunity to address the TODO in `workspace/data-proxy/src/utils/replace-params.ts` and check the ENV var replacement during config parsing.

Thanks.. addressed in https://github.com/sedaprotocol/seda-data-proxy/pull/61/commits/0848a66a1054876e03ff20dbb3818fda915d3c61.
(hope you don't find it too messy but please let me know)

hacheigriega:
> This is maybe also a good opportunity to address the TODO in `workspace/data-proxy/src/utils/replace-params.ts` and check the ENV var replacement during config parsing.

I think the TODO you mentioned is already being done. But should we throw an error in case the env var is not available due to some unexpected reason?

hacheigriega:
> Maybe I'm missing something but I don't get a warning about the env var missing with the following config:
> 
> ```json
> {
> 	"routeGroup": "proxy",
> 	"routes": [
> 		{
> 			"path": "mock/:targets",
> 			"upstreamUrl": "http://3.8.215.180:8080/api/{$THIS_SHOULD_NOT_EXIST}/{:target}",
> 			"method": ["GET"]
> 		}
> 	]
> }
> ```

Ah, I didn't realize the upstream URL could also contain an environment variable syntax. But as I am adding it, I am wondering if it makes sense to have an environment variable in the upstream URL. It also looks like in `proxy-server.ts` the upstream URL is replaced only by the params provided by the router. So now I am thinking we should just emit an error if we detect an environment variable syntax in the upstream URL?

Thomasvdam:
> > Maybe I'm missing something but I don't get a warning about the env var missing with the following config:
> > ```json
> > {
> > 	"routeGroup": "proxy",
> > 	"routes": [
> > 		{
> > 			"path": "mock/:targets",
> > 			"upstreamUrl": "http://3.8.215.180:8080/api/{$THIS_SHOULD_NOT_EXIST}/{:target}",
> > 			"method": ["GET"]
> > 		}
> > 	]
> > }
> > ```
> 
> Ah, I didn't realize the upstream URL could also contain an environment variable syntax. But as I am adding it, I am wondering if it makes sense to have an environment variable in the upstream URL. It also looks like in `proxy-server.ts` the upstream URL is replaced only by the params provided by the router. So now I am thinking we should just emit an error if we detect an environment variable syntax in the upstream URL? (and only allow environment variable syntax in statusEndpoints.apiKey or headers)

The environment variable syntax was meant for the upstream URL. If the upstream contains an API key in the path for example we would prefer to inject it through an env var rather than having the secret in the JSON.

hacheigriega:
I think we can also close #26 as we can run it in silent mode to get the public key?:
```
❯ bun run start validate --silent
$ bun run check-ts && bun run ./workspace/data-proxy validate --silent
$ bunx tsc --noEmit
2025-06-20 14:48:56.830 info: Using config: ./config.json
🌐 Network: testnet

🔐 Using public key (registration verified ✅): 02f2f92d36ad5e6613c1df210ba9a68272375926a605b716b4cd30d49af9910129

✅ SEDA Data Proxy configuration is valid
```

PR #64, opened 2025-07-23
Chore/add network config

mariocao:
## Motivation

When deploying a data proxy with a docker image to Mainnet, we realized that it could be easier to have a config parameter in the key file. That should also help avoiding reusing same keys for different environments.

## Explanation of Changes

- Add optional network field in the key file.
   - If not present, testnet will be used.
   - Same behaviour as previous version.
- Modify `init`, `register` and `run` commands so that they read the file
- Add util function to read network from file (there is potential refactor in here because usually the priv key is read from same file).

## Testing

Checked out all commands.

### DIFF
--- a/workspace/data-proxy/src/cli/utils/private-key.ts
+++ b/workspace/data-proxy/src/cli/utils/private-key.ts
@@ -1,4 +1,5 @@
 import { readFile } from "node:fs/promises";
+import type { Environment } from "@seda-protocol/data-proxy-sdk";
 import { tryAsync, tryParseSync, trySync } from "@seda-protocol/utils";
 import { Result } from "true-myth";
 import {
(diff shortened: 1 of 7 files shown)


workspace/data-proxy/src/cli/utils/private-key.ts
```diff
@@ -15,6 +16,46 @@ async function readPrivateKeyFile(
...
+			resultError += `${error.message} on config property "${error.path?.[0].key}" \n`;
+		}
+
+		return Result.err(`Failed to parse private key file: \n ${resultError}`);
+	}
+
+	return Result.ok(parsedPrivateKeyFile.value.network);
+}
+
 export async function loadPrivateKey(
```
mariocao, line 59: This is the part that is potentially refactorizable.
I did not do it because of time pressure and I didn't want to touch additional code.
Happy if anyone wants to do it.

FranklinWaller (approved):

PR #98, opened 2026-03-25
fix(health): fix issue where chain id wasn't cached

FranklinWaller:
## Motivation

(Write your motivation here)

## Explanation of Changes

(Write your explanation here)

## Testing

(Write your test plan here)

## Related PRs and Issues

(Link your related PRs and Issues here)

### DIFF
--- a/workspace/data-proxy/src/status-plugin/status-plugin.ts
+++ b/workspace/data-proxy/src/status-plugin/status-plugin.ts
@@ -1,5 +1,8 @@
 import type { DataProxy } from "@seda-protocol/data-proxy-sdk";
+import { Duration, Match, Option, Runtime } from "effect";
+import { Effect } from "effect";
 import Elysia from "elysia";
+import { Result } from "true-myth";
 import type { Config } from "../config-parser";
 import logger from "../logger";
 import { getRpcChainId } from "../services/get-rpc-chain-id";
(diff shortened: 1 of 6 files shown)


kaynetik (approved):

Thomasvdam:
I'm not sure if this entirely correct, because now the health check doesn't actually verify that the proxy can reach the chain in future health checks. Maybe wrap it in a cache with TTL until we have time to properly think of how to do a health check mechanism that harmonises core and Fast?

FranklinWaller:
@Thomasvdam I changed it to do without caching because we do indeed want to know which proxies are unhealthy. Instead i've opted for just the fastOnly flag. Then the health endpoint won't return a unhealthy status when the chain is upgrading.

FranklinWaller:
Also made the better effect version, this is just a hotfix

https://github.com/sedaprotocol/seda-data-proxy/pull/100

Thomasvdam (approved):

workspace/data-proxy/src/status-plugin/status-plugin.ts
```diff
@@ -1,7 +1,10 @@
 import type { DataProxy } from "@seda-protocol/data-proxy-sdk";
-import { Effect, Runtime } from "effect";
+import { Duration, Match, Option, Runtime } from "effect";
+import { Effect } from "effect";
```
mariocao, line 3:
```suggestion
import { Duration, Effect, Match, Option, Runtime } from "effect";
```

workspace/data-proxy/src/status-plugin/status-plugin.ts
```diff
@@ -11,71 +14,79 @@ export const statusPlugin = (
...
+						group.get("health", async ({ set }) => {
+							let healthy = true;
 
-					return Response.json({
-						status: chainId.isOk && hasCorrectChainId ? "healthy" : "unhealthy",
-						metrics: context.getMetrics(),
-					});
-				});
+							if (config.fastOnly) {
+								healthy = true;
```
mariocao, line 52: This is redundant (line 49)

workspace/data-proxy/src/status-plugin/status-plugin.ts
```diff
@@ -11,71 +14,79 @@ export const statusPlugin = (
...
+								status: healthy ? "healthy" : "unhealthy",
+								metrics: context.getMetrics(),
+							});
+						});
 
-				return group;
-			});
+						group.get("info", async () => {
+							const chainId = config.fastOnly
+								? Result.ok(dataProxy.options.chainId)
```
mariocao, line 71: should this not be `null`?
Just to avoid that if it is `fastOnly`, it is clear that chainId was not being used (even though it might be set).

workspace/data-proxy/src/status-plugin/status-plugin.ts
```diff
@@ -11,71 +14,79 @@ export const statusPlugin = (
...
+						return group;
+					});
+
+					logger.info(
+						`Status endpoints: /${options.root}/health, /${options.root}/info`,
+					);
+					return app.use(plugin);
+				}),
+			);
 	});
```
mariocao, line 92: I think that if we add the runtime as an expected input param, we could simplify a little bit the `statusPlugin` code.

On the proxy-server.ts:
```
// Get the runtime once at startup
const runtime = Effect.runSync(Effect.runtime<never>());

server.use(
	statusPlugin(statusContext, dataProxy, config.statusEndpoints, config, runtime),
);
```

Entirely optional, we could avoid the outer Effect.gen:
```
export const statusPlugin = (
    context: Context,
    dataProxy: DataProxy,
    options: Config["statusEndpoints"],
    config: Config,
    runtime: Runtime.Runtime<never>,   // ← accept runtime as param
) =>
    (app: Elysia) => {
        const plugin = new Elysia({ name: "status" });

        plugin.group(options.root, (group) => {
            // ... all the route handlers, using runtime directly
        });

        logger.info(`Status endpoints: /${options.root}/health, /${options.root}/info`);
        return app.use(plugin);
    };
```

mariocao (approved):
As I was told, you can forget about my comment regarding the Effect.gen as this is more a PR for a fix. We can improve that in another PR later.

PR #104, opened 2026-04-28
Improve upstream error logging

### LINKED ISSUE #94: ✨ Improve upstream error handling
## ✨ Feature

Improve upstream error handling by logging more context from upstream responses (e.g. request URL, request body, and relevant response details) to make debugging easier.

## Motivation

We recently ran into an issue where upstream calls were failing but the logs did not provide enough information to understand what was happening. This made it difficult to diagnose the root cause.

Having more visibility into failing upstream requests would significantly improve debugging. In particular, it would be useful to log the request URL, request body, and relevant response information when an upstream error occurs.

At the same time, we need to ensure sensitive information such as API keys is not leaked in logs. Some upstream providers unfortunately include API keys directly in query parameters instead of headers, so any solution should account for this.

## Implementation

Extend upstream error logging to include additional request/response context for failing calls.

Sensitive information should be redacted before logging (e.g. API keys in query parameters or request bodies). This could be handled via configurable redaction rules or patterns defined in the config.

Optionally, this logging could be controlled through a configuration parameter so operators can enable or disable extended upstream error logging depending on their needs.

hacheigriega:
## Explanation of Changes

- Adds more context to upstream error logs.
- For JSON path error (`QueryJsonError`), we now also log the full JSON data. => We may want to impose some limit on data size here.
- Also fixes the bug where not all sensitive values from environment variables were redacted.

When an upstream HTTP fetch fails:
```
2026-04-28T17:57:16.820Z ERROR #5 [/proxy/random2] [30a9f2a7-5972-4bf8-a7b5-0983af5f4705] [GET] {
  "message": "Upstream HTTP request failed",
  "routePath": "/proxy/random2",
  "method": "GET",
  "clientRequestUrl": "http://localhost:5384/proxy/random2",
  "upstreamRequestUrl": "https://invalid.com/test?api_key=<redacted>",
  "upstreamRequestHeaders": {
    "user-agent": "curl/8.7.1",
    "accept": "*/*",
    "content-type": "application/json",
    "authorization": "Bearer <redacted>"
  }
}
2026-04-28T17:57:16.821Z ERROR #5 [/proxy/random2] [30a9f2a7-5972-4bf8-a7b5-0983af5f4705] [GET] {
  "error": {
    "error": {
      "error": {
        "code": "ConnectionRefused",
        "path": "https://invalid.com/test?api_key=<redacted>",
        "errno": 0
      },
      "_tag": "HttpClientRequestFailedError",
      "message": "HTTP client request failed: Error: Unable to connect. Is the computer able to access the url?"
    },
    "routePath": "/proxy/random2",
    "_tag": "UpstreamRequestFailedError",
    "message": "Upstream request failed for route /proxy/random2: HttpClientRequestFailedError: HTTP client request failed: Error: Unable to connect. Is the computer able to access the url?"
  },
  "headers": {
    "host": "localhost:5384",
    "user-agent": "curl/8.7.1",
    "accept": "*/*"
  },
  "path": "/proxy/random2",
  "method": "GET",
  "requestUrl": "http://localhost:5384/proxy/random2"
}
```

## Related PRs and Issues

Closes: #94

### DIFF
--- a/workspace/data-proxy/src/controllers/proxy/handle-proxy-request.ts
+++ b/workspace/data-proxy/src/controllers/proxy/handle-proxy-request.ts
@@ -5,6 +5,7 @@ import { JSON_PATH_HEADER_KEY } from "../../constants";
 import {
 	FailedToParseResponseBodyError,
 	NotOkUpstreamResponseError,
+	QueryJsonError,
 	UpstreamRequestFailedError,
 } from "../../errors";
 import { ModuleService } from "../../modules/module";
@@ -12,7 +13,7 @@ import type { ProxyServerOptions } from "../../proxy-server";
 import { HttpClientService } from "../../services/http-client";
 import { createSignedResponseHeaders } from "../../utils/create-headers";
 import { maybeToOption } from "../../utils/effect-utils";
-import { QueryJsonError, queryJson } from "../../utils/query-json";
+import { queryJson } from "../../utils/query-json";
 import { replaceParams } from "../../utils/replace-params";
 import { createUrlSearchParams } from "../../utils/search-params";
 import { injectSearchParamsInUrl } from "../../utils/url";
@@ -213,6 +214,7 @@ export const handleProxyRequest = (inputParams: HandleProxyRequestParams) =>
 					(error) =>
 						new QueryJsonError({
 							error: error.message,
+							data: error.data,
 							type: "config",
 							status: 500,
 						}),
@@ -243,6 +245,7 @@ export const handleProxyRequest = (inputParams: HandleProxyRequestParams) =>
 					(error) =>
 						new QueryJsonError({
 							error: error.message,
+							data: error.data,
 							type: "header",
 							// Fault is from the user side
 							status: 400,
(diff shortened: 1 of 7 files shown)


workspace/data-proxy/src/controllers/proxy/handle-proxy-request.ts
```diff
@@ -128,6 +130,18 @@ export const handleProxyRequest = (inputParams: HandleProxyRequestParams) =>
 							body: Option.getOrUndefined(body),
 						})
 						.pipe(
+							Effect.tapError((error) =>
+								Effect.logError({
+									message: "Upstream HTTP request failed",
```
Thomasvdam: Just curious, is there a benefit to having message as a keyValue over `logError("message_goes_here", { object_with_things });`?

workspace/data-proxy/src/logger.ts
```diff
@@ -24,14 +24,14 @@ import {
 
 let envSecrets: Set<string> = new Set();
 
-function redactEnvSecrets(message: string | unknown) {
+export function redactEnvSecrets(message: string | unknown) {
```
Thomasvdam: I don't think we need to export this?

workspace/data-proxy/src/errors.ts
```diff
@@ -45,3 +45,12 @@ export class FailedToParseConfigError extends Data.TaggedError(
 )<{ error: string | unknown }> {
 	message = `Failed to parse config: ${this.error}`;
 }
+
+export class QueryJsonError extends Data.TaggedError("QueryJsonError")<{
+	error: string | unknown;
+	data?: string;
```
Thomasvdam, line 51: Did we check that this doesn't leak the full response to the caller?

workspace/data-proxy/src/logger.ts
hacheigriega: Remnant from when I was debugging.. thanks, removed

workspace/data-proxy/src/errors.ts
hacheigriega: Thanks for catching this. I forgot to add back the code for clearing the field when I switched the strategy. Please check out the fixup commit

Thomasvdam: Maybe it's a good idea to add an integration test for this? At least when it is applied from the proxy config, because that's also a way to prevent data from reaching the caller.

workspace/data-proxy/src/controllers/proxy/handle-proxy-request.ts
hacheigriega: Nope, changed it to match the other logging pattern

workspace/data-proxy/src/errors.ts
hacheigriega: Yeah good idea, added.

Thomasvdam: Now I'm being a little annoying, but what if the sensitive data happens to be in the first few characters of the stringified response? I think when the config JSON path fails we probably shouldn't return the raw input in any form

hacheigriega: Brilliant.. thanks again. Added fixup commit
Will also double check other existing places for possible leak

Thomasvdam: Sorry I wasn't clear D:

I meant what if the upstream responds with something like:
```json
{
  "key": "value_I_don't_want_to_expose_but_is_not_configure_as_a_secret_I_only_try_to_hide_it_with_the_json_path",
  "data_intended_for_client": "this_should_be_selected_with_jsonpath"
}
```

hacheigriega: Oh, so just we can just not return any data from upstream when JSON Path fails, right? 
Updated. Thanks again again

Thomasvdam: I think that's the safest bet for the case where we apply the config provided JSON path. For a user provided JSON path we can return the data as that should already be 'safe' to consume according to the config.

hacheigriega: Oops, can you approve if you forgot?

Thomasvdam: Sure! Maybe we can create an issue to return the what is left of the upstream data after the operator supplied JSON path when the user supplied JSON path fails?

Thomasvdam (approved):

workspace/data-proxy/src/errors.ts
hacheigriega: Are you suggesting that we simply trust the result of the operator-supplied JSON path? (That it won't contain sensitive data?)

Thomasvdam: I think that's true? From my understanding they have different purposes:

**Operator supplied**: limits what the caller receives. One potential usecase was so you can have 2 data proxies with different fees exposing different data from the same upstream endpoint. The goal here is to limit data access, which is why I think we should not leak information about this data.

**User supplied**: limits the amount of bytes returned by the data proxy. If this is not supplied by the user the proxy will return everything, which can be waay too much for some APIs like the Hyperliquid one for example. The goal here is to save gas on the OP side. If it's not supplied the caller receives all data anyway, so it any data in it should be ok to return when the JSON path fails. Of course provided that this is on the output of the operator supplied JSON path, not the raw upstream data.

hacheigriega: Thanks for the explanation.. True, I see the key point is that the caller can receive all data anyways by omitting json path in the request. Added a commit here to address this.

Thomasvdam (approved):

PR #105, opened 2026-04-29
Clean up waiters when prices get deleted or fail

Thomasvdam:
## Motivation

Fixes a (very minor) leak where old waiters were not properly cleaned up.

## Explanation of Changes

We never `yielded` the `SynchronizedRef.update`, which meant it never got executed. Now we return it from `deletePrice` which already got yielded at the callsite.

I think it makes sense to clear the waiters after a price is resolved or failed, but open to be challenged here.

## Testing

Added failing tests that pass with the fix.

## Related PRs and Issues

N.A.

### DIFF
--- a/workspace/data-proxy/src/modules/pyth-lazer/price-cache.test.ts
+++ b/workspace/data-proxy/src/modules/pyth-lazer/price-cache.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "bun:test";
-import { Effect, Fiber } from "effect";
+import { Effect, Either, Fiber } from "effect";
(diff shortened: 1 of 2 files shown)


hacheigriega (approved):
Not so minor it seems? :D And wouldn't it be incorrect to leave the waiter uncleared in `setPriceToError` because no new waiter would be created afterwards?

Thomasvdam:
>And wouldn't it be incorrect to leave the waiter uncleared in setPriceToError because no new waiter would be created afterwards?

I think so, but wasn't 100% confident :)

PR #107, opened 2026-04-30
feat(hydromancer): add WS-first module with REST fallback

jaspersagent:
Adds a `hydromancer` module type. The WS connection holds a per-coin `activeAssetCtx` cache; the first request for a coin enqueues a subscribe and is served via the batch REST endpoint until the cache warms, then subsequent requests within the staleness window serve from memory. When the WS is down or the cached frame is stale, the request falls through to REST. Response shape is `{coin: ctx}`, matching the upstream batch endpoint and the existing passthrough route, so callers can swap between the two without reshaping.

- Demand-driven subscribe pattern mirrors `pyth-lazer` (Queue + forkDaemon, single subscribe call site). Idle cleanup is also factored into a shared `forkIdleCleanup` helper used by both modules.

### DIFF
--- a/workspace/data-proxy/src/modules/hydromancer/asset-cache.ts
+++ b/workspace/data-proxy/src/modules/hydromancer/asset-cache.ts
@@ -1,0 +1,43 @@
+import { Effect, MutableHashMap, Option } from "effect";
+import type { AssetCtx } from "../../config/hydromancer-module-config";
+
+export interface AssetCacheEntry {
+	ctx: AssetCtx;
+	lastUpdate: number;
+}
+
+export const createAssetCache = () =>
+	Effect.gen(function* () {
+		const entries = MutableHashMap.empty<string, AssetCacheEntry>();
+
+		const set = (coin: string, ctx: AssetCtx, now: number) =>
+			Effect.sync(() => {
+				MutableHashMap.set(entries, coin, { ctx, lastUpdate: now });
+			});
+
+		const get = (coin: string) =>
+			Effect.sync(() => MutableHashMap.get(entries, coin));
+
+		const remove = (coin: string) =>
+			Effect.sync(() => {
+				MutableHashMap.remove(entries, coin);
+			});
+
+		const isFresh = (coin: string, staleAfterMillis: number, now: number) =>
+			Effect.sync(() => {
+				const entry = MutableHashMap.get(entries, coin);
+				if (Option.isNone(entry)) return false;
+				return now - entry.value.lastUpdate <= staleAfterMillis;
+			});
+
+		return {
+			set,
+			get,
+			remove,
+			isFresh,
+		};
+	});
+
+export type AssetCache = Effect.Effect.Success<
+	ReturnType<typeof createAssetCache>
+>;
--- a/workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
+++ b/workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
@@ -1,0 +1,136 @@
+import { Clock, Duration, Effect, Layer, MutableHashMap, Option } from "effect";
+import type { Route } from "../../config/config-parser";
+import type {
+	AssetCtx,
+	HydromancerModuleConfig,
+} from "../../config/hydromancer-module-config";
+import { createErrorResponse } from "../../controllers/create-error-response";
+import { forkIdleCleanup } from "../../utils/idle-cleanup";
+import { replaceParams } from "../../utils/replace-params";
+import { FailedToHandleRequest, ModuleService } from "../module";
+import { createAssetCache } from "./asset-cache";
+import { FailedToHandleHydromancerRequestError } from "./errors";
+import { fetchAssetContextsFromRest } from "./rest-fallback";
+import { createHydromancerWS } from "./ws-client";
+
+export const HydromancerModuleService = (config: HydromancerModuleConfig) =>
+	Layer.effect(
+		ModuleService,
+		Effect.gen(function* () {
+			yield* Effect.logInfo("Initializing Hydromancer module", {
+				name: config.name,
+				wsUrl: config.wsUrl,
+				restBaseUrl: config.restBaseUrl,
+			});
+
+			const cache = yield* createAssetCache();
+			const ws = yield* createHydromancerWS(config, cache);
+			const staleAfterMillis = Duration.toMillis(config.staleAfter);
+			const lastRequestToCoin = MutableHashMap.empty<string, number>();
+
+			const start = () =>
+				Effect.gen(function* () {
+					yield* Effect.logInfo("Hydromancer module started", {
+						name: config.name,
+					});
+
+					yield* ws.start();
+
+					for (const coin of config.subscriptionCoins) {
+						yield* ws.subscribe(coin);
+					}
+
+					yield* forkIdleCleanup({
+						lastRequest: lastRequestToCoin,
+						ttl: config.coinsCleanupTtl,
+						interval: config.coinsCleanupInterval,
+						onExpire: (coin) =>
+							Effect.gen(function* () {
+								yield* Effect.logInfo("Cleaning up idle coin", { coin });
+								yield* cache.remove(coin);
+								yield* ws.unsubscribe(coin);
+							}),
+					});
+				}).pipe(Effect.annotateLogs("_name", "hydromancer"));
+
+			const handleRequest = (
+				route: Route,
+				params: Record<string, string>,
+				_request: Request,
+			) =>
+				Effect.gen(function* () {
+					if (route.type !== "hydromancer") {
+						return yield* Effect.fail(
+							new FailedToHandleRequest({
+								msg: "Route is not a Hydromancer module",
+							}),
+						);
+					}
+
+					const coins = replaceParams(route.fetchFromModule, params).split(",");
+
+					if (coins.length > config.maxCoinsPerRequest) {
+						return yield* Effect.fail(
+							new FailedToHandleHydromancerRequestError({
+								error: `Too many coins, max is ${config.maxCoinsPerRequest} but got ${coins.length}`,
+								status: 400,
+							}),
+						);
+					}
+
+					const now = yield* Clock.currentTimeMillis;
+					const socketHealthy = !(yield* ws.hasError());
+
+					for (const coin of coins) {
+						yield* ws.subscribe(coin);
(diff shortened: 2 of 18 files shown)


workspace/data-proxy/src/utils/idle-cleanup.ts
```diff
@@ -0,0 +1,50 @@
...
+export const forkIdleCleanup = <K>(
+	options: ForkIdleCleanupOptions<K>,
+): Effect.Effect<Fiber.RuntimeFiber<unknown, unknown>, never, never> =>
+	Effect.forkDaemon(
+		Effect.gen(function* () {
+			const now = yield* Clock.currentTimeMillis;
+			const ttlMs = Duration.toMillis(options.ttl);
+			for (const [key, lastTs] of MutableHashMap.fromIterable(
+				options.lastRequest,
+			)) {
```
Thomasvdam: This `fromIterable` is redundant. It's been fixed in the upstream, but in the original file.

workspace/data-proxy/src/utils/idle-cleanup.test.ts
```diff
@@ -0,0 +1,290 @@
...
+				// Even though we advance well past TTL, the daemon is dead.
+				yield* TestClock.adjust(Duration.seconds(600));
+
+				expect(Option.isSome(MutableHashMap.get(lastRequest, "BTC"))).toBe(
+					true,
+				);
+				expect(expired).toEqual([]);
+			}),
+		);
+	});
```
Thomasvdam: This test doesn't add anything, it just tests Effect SDK behaviour.

workspace/data-proxy/src/utils/idle-cleanup.ts
```diff
@@ -0,0 +1,50 @@
...
+				if (now - lastTs > ttlMs) {
+					MutableHashMap.remove(options.lastRequest, key);
+					yield* options
+						.onExpire(key)
+						.pipe(
+							Effect.catchAll((error) =>
+								Effect.logWarning("idle-cleanup onExpire failed", { error }),
+							),
+						);
+				}
```
Thomasvdam: I'm not sure if this makes sense? Now the idle-cleanup will never be called again for this key, there's something to be said to keeping the entry if the cleanup fails so it gets retried.

workspace/data-proxy/src/utils/idle-cleanup.test.ts
```diff
@@ -0,0 +1,290 @@
...
+
+				yield* TestClock.adjust(Duration.seconds(600));
+
+				expect(MutableHashMap.size(lastRequest)).toBe(0);
+				expect(expired).toEqual([]);
+
+				yield* Fiber.interrupt(fiber);
+			}),
+		);
+	});
```
Thomasvdam: This test is kinda useless. I think it's better if we test that the expire doesn't get called when the map is empty.

```ts
onExpire: (key) => Effect.die("Unexpected call to onExpire")
```

workspace/data-proxy/src/utils/idle-cleanup.test.ts
```diff
@@ -0,0 +1,290 @@
...
+				);
+				expect(Option.isSome(MutableHashMap.get(lastRequest, "FRESH"))).toBe(
+					true,
+				);
+				expect(expired).toEqual(["OLD"]);
+
+				yield* Fiber.interrupt(fiber);
+			}),
+		);
+	});
```
Thomasvdam: I see this pattern a lot where it claims it's testing with multiple keys, but in reality it only adds 1 key, advances time, then adds another key, advances time, and checks.

It should add both keys at start of the test with different timestamps, then advance time (optionally in stages) and verify results.

workspace/data-proxy/src/modules/hydromancer/ws-client.ts
```diff
@@ -0,0 +1,162 @@
...
+		return null;
+	}
+	return parsed.value.data;
+};
+
+const defaultReconnectSchedule = (config: HydromancerModuleConfig) =>
+	Schedule.exponential(Duration.seconds(1)).pipe(
+		Schedule.either(Schedule.spaced(config.reconnectMaxBackoff)),
+	);
+
```
Thomasvdam: If I'm not mistaken this schedule never resets. So even if the failures are days apart it will still advance until all retries take `reconnectMaxBackoff` time.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -0,0 +1,171 @@
...
+				wsUrl: config.wsUrl,
+				restBaseUrl: config.restBaseUrl,
+			});
+
+			const cache = yield* createAssetCache();
+			const staleAfterMillis = Duration.toMillis(config.staleAfter);
+			const newCoinQueue = yield* Queue.unbounded<string>();
+			const lastRequestToCoin = MutableHashMap.empty<string, number>();
+			const desiredCoins = MutableHashMap.empty<string, true>();
+			const currentWS: { value: WebSocket | null } = { value: null };
```
Thomasvdam: This is a really ugly way to pass the reference around :P The WS should live in a service that exposes methods to interact with it.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+			const start = () =>
+				Effect.gen(function* () {
+					yield* Effect.logInfo("Hydromancer module started", {
+						name: config.name,
+					});
+
+					for (const coin of config.subscriptionCoins) {
+						yield* Queue.offer(newCoinQueue, coin);
```
Thomasvdam: I think this suffers from the same cleanup leak I fixed in the Pyth module.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -0,0 +1,171 @@
...
+							const ctx = restBatch[coin];
+							if (ctx) {
+								yield* cache.set(coin, ctx, now);
+							}
+						}
+					}
+
+					const resolved: Record<string, AssetCtx> = {};
+					for (const coin of coins) {
+						const ctx = cached.get(coin) ?? restBatch[coin];
```
Thomasvdam: This is redundant, if it is in `restBatch` it is also in the cache.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -0,0 +1,171 @@
...
+							}),
+						);
+					}
+
+					const now = yield* Clock.currentTimeMillis;
+
+					for (const coin of coins) {
+						if (Option.isNone(MutableHashMap.get(lastRequestToCoin, coin))) {
+							yield* Queue.offer(newCoinQueue, coin);
+						}
```
Thomasvdam: Multiple coins could be offered, we should check somewhere that the coin is not already tracked.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+					yield* startWebSocketDaemon(config, cache, desiredCoins, currentWS);
+
+					yield* Effect.forkDaemon(
+						Effect.gen(function* () {
+							const coin = yield* Queue.take(newCoinQueue);
+							MutableHashMap.set(desiredCoins, coin, true);
+							const ws = currentWS.value;
+							if (ws !== null && ws.readyState === WebSocket.OPEN) {
+								ws.send(buildSubscribeFrame(coin));
```
Thomasvdam: If `send` throws an error the daemon does not recover, but this should probably be addressed by abstracting the WS.

workspace/data-proxy/src/modules/hydromancer/rest-fallback.ts
```diff
@@ -0,0 +1,87 @@
+import { tryParseSync } from "@seda-protocol/utils";
+import { Effect } from "effect";
+import * as v from "valibot";
+import {
+	AssetCtxSchema,
+	type HydromancerModuleConfig,
+} from "../../config/hydromancer-module-config";
+import { FailedToHandleHydromancerRequestError } from "./errors";
+
+const FETCH_TIMEOUT_MS = 15_000;
```
Thomasvdam: This should probably be a config variable

workspace/data-proxy/src/modules/hydromancer/asset-cache.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface AssetCacheEntry {
+	ctx: AssetCtx;
+	lastUpdate: number;
+}
+
+export const createAssetCache = () =>
+	Effect.gen(function* () {
+		const entries = MutableHashMap.empty<string, AssetCacheEntry>();
+		let socketError: string | undefined;
```
Thomasvdam: This is not used for anything

Thomasvdam (changes requested):

workspace/data-proxy/src/modules/hydromancer/ws-client.ts
```diff
@@ -0,0 +1,218 @@
...
+							error: String(err),
+						}),
+					);
+					try {
+						ws.close();
+					} catch {
+						// best-effort; the close listener will trigger the reconnect loop.
+					}
+				}
+			});
```
Thomasvdam: Why is this an `Effect.sync` if we're also using an Effect inside for the logging?

workspace/data-proxy/src/modules/hydromancer/ws-client.ts
```diff
@@ -0,0 +1,218 @@
...
+				Effect.sync(() => {
+					socketError = `ws connect failed: ${String(error)}`;
+				}),
+			),
+			Effect.retry(schedule),
+		);
+
+		const start = () => Effect.forkDaemon(loop);
+
+		return { start, subscribe, unsubscribe, hasError } satisfies HydromancerWS;
```
Thomasvdam: Obviously user error if it happens, but if someone calls `start` multiple times we end with unreachable WS instances.

Thomasvdam (commented):
It's still behind the upstream. Leaving final review to HY

hacheigriega:
> It's still behind the upstream. Leaving final review to HY

Didn't find anything problematic beyond those two good points you made. I suppose we can just get rid of start() API?

Thomasvdam (approved):

PR #108, opened 2026-05-01
LO:TECH WebSocket module

hacheigriega:
## Motivation

Implement LO:TECH module based on WebSocket connection. 
The logic is largely taken from the existing Pyth-Lazer module, but unlike Pyth-Lazer that uses a client library, this module uses the ws WebSocket package directly.

## Explanation of Changes

The API key is available on 1Pass. 
Currently, the API key only gives access to the "PRICE" data type for the following symbols: NVDA, WTIK6, WTIM6, WTIN6, WTIQ6, and WTIU6. The current implementation also only supports the "PRICE" data type, but this can be extended in the future. 

Example configuration:
```json
{
	"modules": [
		{
			"type": "lo-tech",
			"name": "lotech",
			"exchange": "rwa",
			"loTechApiKeyEnvKey": "LOTECH_API_KEY",
			"priceFeeds": [
				{
					"symbol": "NVDA",
					"dataType": "PRICE"
				}
			]
		}
	],
	"routes": [
		{
			"type": "lo-tech",
			"moduleName": "lotech",
			"path": "/lotech/:streams",
			"method": "GET",
			"fetchFromModule": "{:streams}"
		}
	]
}
```

After running the data proxy with the command
```bash
bun start run --disable-proof -c lo-tech.json
```

query the symbols through curl:
```bash
curl http://127.0.0.1:5384/proxy/lotech/NVDA,WTIK6,WTIM6,WTIN6,WTIQ6,WTIU6
```

## Testing

(Write your test plan here)

### DIFF
--- a/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
+++ b/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
@@ -1,0 +1,225 @@
+import {
+	Clock,
+	Duration,
+	Effect,
+	Layer,
+	Match,
+	MutableHashMap,
+	Option,
+	Queue,
+	Schedule,
+} from "effect";
+import type { Route } from "../../config/config-parser";
+import {
+	LO_TECH_DATA_TYPE_PRICE,
+	type LoTechModuleConfig,
+} from "../../config/lo-tech-module-config";
+import { createErrorResponse } from "../../controllers/create-error-response";
+import { replaceParams } from "../../utils/replace-params";
+import { FailedToHandleRequest, ModuleService } from "../module";
+import { FailedToHandleLoTechRequestError } from "./errors";
+import { createPriceCache } from "./price-cache";
+import type { LoTechDataPrice, LoTechParsedData } from "./schema";
+import { makeLoTechWebSocketService } from "./ws-client";
+
+export type PriceFeedSymbol = string;
+
+export const LoTechModuleService = (config: LoTechModuleConfig) =>
+	Layer.effect(
+		ModuleService,
+		Effect.gen(function* () {
+			yield* Effect.logInfo("Initializing LO:TECH module");
+
+			const runtime = yield* Effect.runtime();
+			const priceCache = yield* createPriceCache();
+			const priceFeeds = MutableHashMap.empty<PriceFeedSymbol, number>();
+			const newPriceFeedRequests = yield* Queue.unbounded<PriceFeedSymbol>();
+			// The timestamp of the last request to the price feed used for cleanup.
+			const lastRequestToPriceFeed = MutableHashMap.empty<
+				PriceFeedSymbol,
+				number
+			>();
+
+			let priceFeedId = 0;
+
+			const updatePrice = (data: LoTechDataPrice) =>
+				Effect.gen(function* () {
+					yield* Effect.logTrace("Received message from LO:TECH client", data);
+
+					// Set the price if the symbol is subscribed to.
+					const { symbol } = data;
+					if (!MutableHashMap.has(priceFeeds, symbol)) {
+						return;
+					}
+					yield* priceCache.setPrice(symbol, data);
+				});
+
+			const handleDataMessage = (data: LoTechParsedData) =>
+				Match.value(data).pipe(
+					Match.discriminatorsExhaustive("type")({
+						[LO_TECH_DATA_TYPE_PRICE]: (priceData) => updatePrice(priceData),
+					}),
+				);
+
+			const loTechWs = yield* makeLoTechWebSocketService({
+				config,
+				runtime,
+				onConnected: (api) =>
+					Effect.gen(function* () {
+						for (const priceFeed of config.priceFeeds) {
+							yield* Effect.logInfo(
+								`Subscribing to price feed ${priceFeed.symbol}`,
+							);
+
+							const newSubscriptionId = priceFeedId++;
+							MutableHashMap.set(
+								priceFeeds,
+								priceFeed.symbol,
+								newSubscriptionId,
+							);
+
+							const now = yield* Clock.currentTimeMillis;
+							MutableHashMap.set(lastRequestToPriceFeed, priceFeed.symbol, now);
+
+							yield* api.subscribePrice(priceFeed.symbol, newSubscriptionId);
+						}
+					}),
+				handleDataMessage,
+			});
+
+			const start = () =>
+				Effect.gen(function* () {
+					yield* Effect.logInfo("Starting LO:TECH module");
+
+					// Background fiber for handling new price feed subscriptions
+					yield* Effect.forkDaemon(
+						Effect.gen(function* () {
+							const newSymbol = yield* newPriceFeedRequests.take;
+
+							yield* Effect.logInfo(`Subscribing to price feed ${newSymbol}`);
+
+							const newSubscriptionId = priceFeedId++;
+							MutableHashMap.set(priceFeeds, newSymbol, newSubscriptionId);
+
+							const now = yield* Clock.currentTimeMillis;
+							MutableHashMap.set(lastRequestToPriceFeed, newSymbol, now);
+
+							yield* loTechWs.subscribePrice(newSymbol, newSubscriptionId);
+						}).pipe(Effect.forever),
+					);
+
+					// Background fiber for cleaning up price feeds subscriptions
+					yield* Effect.forkDaemon(
+						Effect.gen(function* () {
+							const now = yield* Clock.currentTimeMillis;
+							yield* Effect.logDebug(
+								`Cleaning up price feeds (currently running ${priceCache.size()} price feeds)..`,
+							);
+
+							for (const [
+								symbol,
+								lastRequestTimestamp,
+							] of lastRequestToPriceFeed) {
+								const cleanupInterval = Duration.toMillis(
+									config.priceFeedsCleanupTtl,
+								);
+								const timeSinceLastRequest = now - lastRequestTimestamp;
+
+								yield* Effect.logDebug(
+									`Time since last request for price feed ${symbol}: ${Duration.format(Duration.decode(timeSinceLastRequest))}`,
+								);
+
+								if (timeSinceLastRequest > cleanupInterval) {
+									yield* Effect.logInfo(`Cleaning up price feed ${symbol}`);
+									MutableHashMap.remove(lastRequestToPriceFeed, symbol);
+									yield* priceCache.deletePrice(symbol);
+
+									const priceFeedId = MutableHashMap.get(priceFeeds, symbol);
+
+									if (Option.isSome(priceFeedId)) {
+										yield* loTechWs.unsubscribePrice(symbol);
+										MutableHashMap.remove(priceFeeds, symbol);
+									}
+								}
+							}
+						}).pipe(
+							Effect.schedule(
+								Schedule.spaced(config.priceFeedsCleanupInterval),
+							),
+						),
+					);
+				}).pipe(Effect.annotateLogs("_name", "lo-tech"));
+
+			const handleRequest = (
+				route: Route,
+				params: Record<string, string>,
+				request: Request,
+			) =>
+				Effect.gen(function* () {
+					if (route.type !== "lo-tech") {
+						return yield* Effect.fail(
+							new FailedToHandleRequest({
+								msg: "Route is not a LO:TECH module",
+							}),
+						);
+					}
+
+					yield* Effect.logInfo("Handling LO:TECH request", { route, params });
+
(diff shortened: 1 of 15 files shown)


workspace/data-proxy/src/modules/lo-tech/price-cache.ts
Thomasvdam, line 1: Feels like this is something we can generalise right? Since the symbol and price are used only as types they can be passed as generics. Only the error would have to be changed to a shared concrete type, but that can be mapped to a module specific error at the call site if required.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -0,0 +1,237 @@
...
+			const handleDataMessage = (data: LoTechData) =>
+				Effect.gen(function* () {
+					yield* Match.value(data).pipe(
+						Match.when({ type: "PRICE" }, (priceData: LoTechDataPrice) => {
+							return updatePrice(priceData);
+						}),
+						Match.orElse(() =>
+							Effect.logError("Unexpected LO:TECH data message type", data),
+						),
+					);
```
Thomasvdam: Is there a way we can tie the supported message types from the config to this match statement? As in that we get a type error when we add a type in 1 place but not the other.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -0,0 +1,237 @@
...
+				for (const [symbol, priceFeedId] of priceFeeds) {
+					socket.send(
+						JSON.stringify({
+							op: "SUBSCRIBE",
+							topics: [{ symbol, type: "PRICE" }],
+							id: priceFeedId,
+						}),
+					);
+				}
+			};
```
Thomasvdam: This feels a little confusing. How I'm reading it:

1. `start` is called, this enqueues all pricefeeds from the config
2. The queue consumer reads the configured pricefeeds 1 by 1
  a. updates the `priceFeeds` map
  b. attempts to subscribe to the feed (silently does nothing when the connection's not open)
3. If the socket wasn't open this code runs when the socket does open, subscribing to the same feeds since they were added to the map in 2a

If the `open` event fires before items are enqueued it still works since this map will be empty, and the `sendIfOpen` should succeed.


While I think this works it's not very clear from the code. I think it makes more sense to make the WS client be responsible for handling asynchronicity in the opening of the connection and when feeds are requested. A possible solution would be a queue on the WS client side that's gated by a latch, or to expose a latch from the WS module and have that gate calls to `send`.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -0,0 +1,237 @@
...
+							MutableHashMap.set(priceFeeds, newSymbol, newSubscriptionId);
+
+							const now = yield* Clock.currentTimeMillis;
+							MutableHashMap.set(lastRequestToPriceFeed, newSymbol, now);
+
+							yield* loTechWs.sendIfOpen(
+								JSON.stringify({
+									op: "SUBSCRIBE",
+									topics: [{ symbol: newSymbol, type: "PRICE" }],
+									// id: newSubscriptionId, // TODO Handle ack
```
Thomasvdam: Is this something we need to address?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -0,0 +1,237 @@
...
+						prices.push(price);
+					}
+
+					return yield* Effect.succeed(
+						new Response(JSON.stringify(prices), { status: 200 }),
+					);
+				}).pipe(
+					Effect.withSpan("handleLoTechRequest"),
+					Effect.catchAll((error) => {
+						// TODO: Handle error properly
```
Thomasvdam: Do we need more handling?

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
```diff
@@ -0,0 +1,180 @@
...
+	handleDataMessage: (data: LoTechData) => Effect.Effect<void>;
+};
+
+export const makeLoTechWebSocketService = (
+	deps: LoTechWebSocketServiceDeps,
+): Effect.Effect<LoTechWebSocketServiceApi> =>
+	Effect.gen(function* () {
+		const { config, runtime, onOpen, handleDataMessage } = deps;
+
+		let activeSocket: WebSocket | null = null;
```
Thomasvdam, line 47: I think we can use `acquireRelease` for this from Effect?

Thomasvdam: Jaspers agent did something like that in #107. Take that as inspiration, not a template :P

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
```diff
@@ -0,0 +1,180 @@
...
+				const sock = activeSocket;
+				if (sock !== null && sock.readyState === WebSocket.OPEN) {
+					yield* Effect.logDebug("Sending message to LO:TECH", { text });
+					sock.send(text);
+				}
+			});
+
+		const runWebSocketSession = (): Promise<void> =>
+			new Promise((resolve) => {
+				const url = `wss://data.lo.tech/ws/v1/${config.exchange}`;
```
Thomasvdam: The base URL might also be a good config option? Not sure if they have different environments with different URLs.

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
```diff
@@ -0,0 +1,180 @@
...
+							}),
+					}).pipe(
+						Effect.catchAll((error) =>
+							Effect.logError("LO:TECH connection session failed", {
+								error,
+							}),
+						),
+					);
+					yield* Effect.sleep(Duration.millis(config.reconnectDelayMs ?? 1000));
+				}
```
Thomasvdam: Can't we express this with `Effect.retry({ /* options here */ })`

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
```diff
@@ -0,0 +1,180 @@
...
+
+					if (pingTimer !== undefined) {
+						clearInterval(pingTimer);
+					}
+					pingTimer = setInterval(() => {
+						Runtime.runSync(
+							runtime,
+							sendIfOpen(JSON.stringify({ op: "PING" })),
+						);
+					}, LO_TECH_PING_INTERVAL_MS);
```
Thomasvdam, line 138: Doesn't `WS` come with a built-in ping/pong mechanism?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Don't think so - removed it

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
hacheigriega: I thought it had to be in the format as described in the doc `{op: "PING"}`, but I see that `socket.ping();` works as well.

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
hacheigriega: All the APIs in the provided doc had the same base URL.
But yes changed it to config

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: I tried handling ack messages and updating the active subscription map when we receive ack messages, but the problem with that approach is that we cannot deliver price data that arrives before ack messages. So I just removed this for now.

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
hacheigriega: Retry is apparently for an effect that should be retried after fails. So I expressed it using Effect.repeat.

Thomasvdam: True, but it's not failing because we're catching the error with `catchAll` :D 

So we could do something like

```ts
yield* Effect.tryPromise({
	try: () => runWebSocketSession(),
	// No need to wrap the error in a new Error
	catch: (error) => error,
}).pipe(
	Effect.tapError((error) => Effect.logError("LO:TECH connection session failed", error)),
	Effect.retry({ /* config */}),
);
```

The `catchAll` + `retry` is also fine. The main thing to address is that we're wrapping all errors from `runWebSocketSession()` in another `new Error` which is not adding context, so I think we're better off just letting it bubble up directly.

workspace/data-proxy/src/modules/lo-tech/price-cache.ts
hacheigriega: I was actually thinking this too, but how about I address this as a new issue to generalize across dxfeed, pyth-lazer, and this?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: OK good point. The WebSocket client now exposes subscribePrice API and uses `pendingOutbound` queue to address the issue where nothing is done when connection is not open. Does this address everything here?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Is there a way to achieve that and also log an error if there is an error (the else case)?

workspace/data-proxy/src/modules/lo-tech/price-cache.ts
Thomasvdam: Sounds good to me!

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: I have no knowledge of their API and what the ACK actually's used for 😅  I saw the TODO and asked the question, but your explanation makes sense. Possibly something to keep in mind if we experience issues in the future?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: Hmm, I think with the current flow that's not possible.

Basically we're doing 2 things in 1 step:
- Narrowing the payload from an (mostly) unknown object to a typed object
- Handling the payload

A more robust approach would be to separate the two. We can narrow the payload with an actual schema (either valibot or Effect.Schema). That way we can parse the data in `ws-client` and have that log an error when we receive data we're not able to handle.
As a result we have a guarantee that anything passed into the actual handler is something we've asserted to be 'correct' and we can let the Match simply route the message to the correct handler.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: I think so :)

workspace/data-proxy/src/modules/lo-tech/price-cache.ts
hacheigriega: https://github.com/sedaprotocol/seda-data-proxy/issues/111

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Thanks.. addressed in the last fixup commit

workspace/data-proxy/src/modules/lo-tech/ws-client.ts
hacheigriega: So this would require some refactor of the client code to bring session within an Effect scope. To be honest I don't fully understand how much we can gain from this.

Thomasvdam: I'm still developing my understanding of it so I can't yet articulate well why I like it. 😬 

I think that in my mind it helps bridge the gap between Effect<>non-Effect, so non-Effectful code can live in the acquireRelease abstraction and the business logic can remain 'pure' Effect.

And I'm generally not a big fan of `let x: Thing | null = null;` and having some code assign/unassign the variable. Usually it makes the code harder to read because there's nothing stopping us from messing with the variable anywhere in the scope, so you can't reason about what happens to the variable without having read all the code and checking all the lines/branches that touch the variable.

That being said, I haven't worked with it too much so it's entirely possible that it doesn't address these issues, or even if it does that it comes at a greater cost in terms of code complexity, so take it with a bag of salt :)

FranklinWaller (approved):

PR #112, opened 2026-05-08
refactor(data-proxy): shared price data cache

hacheigriega:
## Explanation of Changes

Generalize price caches from the modules dxfeed, lo-tech, and pyth-lazer. We keep the existing caching tests in the modules.


workspace/data-proxy/src/modules/pyth-lazer/price-cache.test.ts
Thomasvdam, line 1: Should this not a test for the shared util?

workspace/data-proxy/src/modules/lo-tech/price-cache.test.ts
Thomasvdam, line 1: Same here

workspace/data-proxy/src/modules/pyth-lazer/price-cache.test.ts
Thomasvdam: Oh I just saw the PR description, but I still think it makes sense to take 1 of the existing test files and just add 1 more test that verifies it works with arbitrary shapes?

hacheigriega: true. addressed in the fixup commit

Thomasvdam (approved):
I think the branches are a little funky, but otherwise good

PR #113, opened 2026-05-08
PM Insights module

hacheigriega:
## Explanation of Changes

Supports PM Insights module by logging in every 50 minutes (required to be refreshed every 1 hour) to obtain an API token.

Example config:
```
{
{
	"modules": [
		{
			"type": "pm-insights",
			"name": "pminsights",
			"baseUrl": "https://api.pminsights.com/",
			"emailEnvKey": "PM_INSIGHTS_EMAIL",
			"passwordEnvKey": "PM_INSIGHTS_PASSWORD",
			"tokenRefreshIntervalMinutes": 50,
			"tokenRetryIntervalMinutes": 5
		}
	],
	"routes": [
		{
			"type": "pm-insights",
			"moduleName": "pminsights",
			"path": "/pminsights/*",
			"upstreamPath": "/{*}",
			"method": ["GET"]
		}
	]
}
```

### DIFF
--- a/workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
+++ b/workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
@@ -1,0 +1,304 @@
+import { layer as fetchHttpClientLayer } from "@effect/platform/FetchHttpClient";
+import * as Headers from "@effect/platform/Headers";
+import { text as httpBodyText } from "@effect/platform/HttpBody";
+import { HttpClient } from "@effect/platform/HttpClient";
+import * as HttpClientRequest from "@effect/platform/HttpClientRequest";
+import { type HttpMethod, hasBody } from "@effect/platform/HttpMethod";
+import { Duration, Effect, Either, Layer, Option, Ref, Schedule } from "effect";
+import type { Route } from "../../config/config-parser";
+import type { PmInsightsModuleConfig } from "../../config/pm-insights-module-config";
+import { createErrorResponse } from "../../controllers/create-error-response";
+import { replaceParams } from "../../utils/replace-params";
+import { FailedToHandleRequest, ModuleService } from "../module";
+import { FailedToHandlePmInsightsRequestError } from "./errors";
+
+const FETCH_TIMEOUT_MS = 15_000;
+
+function parseTokenFromLoginBody(text: string): string | undefined {
+	const trimmed = text.trim();
+	if (trimmed.startsWith("{")) {
+		const jsonEither = Either.try(
+			() => JSON.parse(trimmed) as Record<string, unknown>,
+		);
+		if (Either.isRight(jsonEither)) {
+			const json = jsonEither.right;
+			const t = json.access_token ?? json.token ?? json.accessToken;
+			if (typeof t === "string" && t.length > 0) {
+				return t;
+			}
+		}
+	}
+	if (/^[\w-]+\.[\w-]+\.[\w-]+$/.test(trimmed)) {
+		return trimmed;
+	}
+	return undefined;
+}
+
+function normalizeHttpMethod(method: string): HttpMethod {
+	const u = method.toUpperCase();
+	switch (u) {
+		case "GET":
+		case "POST":
+		case "PUT":
+		case "DELETE":
+		case "PATCH":
+		case "HEAD":
+		case "OPTIONS":
+			return u;
+		default:
+			return "GET";
+	}
+}
+
+export const PmInsightsModuleService = (config: PmInsightsModuleConfig) =>
+	Layer.effect(
+		ModuleService,
+		Effect.gen(function* () {
+			yield* Effect.logInfo("Initializing PM Insights module", {
+				name: config.name,
+				baseUrl: config.baseUrl,
+			});
+
+			const tokenRef = yield* Ref.make<Option.Option<string>>(Option.none());
+
+			const loginUrl = new URL("login", config.baseUrl).toString();
+
+			const performLogin = Effect.gen(function* () {
+				const body = new URLSearchParams({
+					username: config.email,
+					password: config.password,
+				}).toString();
+
+				const client = yield* HttpClient;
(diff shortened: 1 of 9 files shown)


FranklinWaller (approved):

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,248 @@
...
+			}
+		} catch {
+			/* try raw JWT below */
+		}
+	}
+	if (/^[\w-]+\.[\w-]+\.[\w-]+$/.test(trimmed)) {
+		return trimmed;
+	}
+	return undefined;
+}
```
FranklinWaller, line 35: Personally i would prefer using the normal effect try route instead of using try catch, you can use Effect.either to catch the error. That would remove the empty catch

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,248 @@
...
+						const isTimeout =
+							error instanceof Error && error.name === "TimeoutError";
+						return new FailedToHandlePmInsightsRequestError({
+							error: isTimeout
+								? `Login timed out after ${FETCH_TIMEOUT_MS}ms`
+								: `Login request failed: ${error}`,
+							status: isTimeout ? 504 : 502,
+						});
+					},
+				});
```
FranklinWaller: You can use Effect HTTP client, which makes testing and this easier: https://github.com/Effect-TS/effect/blob/main/packages/platform/README.md#http-client

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,248 @@
...
+											err instanceof FailedToHandlePmInsightsRequestError
+												? err.message
+												: `${err}`,
+									},
+								);
+							}
+						}),
+					);
+
+					yield* Effect.forkDaemon(refreshLoop);
```
FranklinWaller: We can make this an effect schedule

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,248 @@
...
+							const isTimeout =
+								error instanceof Error && error.name === "TimeoutError";
+							return new FailedToHandlePmInsightsRequestError({
+								error: isTimeout
+									? `PM Insights request timed out after ${FETCH_TIMEOUT_MS}ms`
+									: `Failed to fetch from PM Insights: ${error}`,
+								status: isTimeout ? 504 : 502,
+							});
+						},
+					});
```
FranklinWaller: Same here, effect client http

FranklinWaller (dismissed):

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,300 @@
...
+					yield* Effect.forkDaemon(
+						Effect.gen(function* () {
+							// Add initial delay before first refresh
+							yield* Effect.sleep(refreshInterval);
+
+							yield* Effect.repeat(
+								refreshLogin,
+								Schedule.spaced(refreshInterval),
+							);
+						}),
```
Thomasvdam: I think you can express this with

```ts
yield* Effect.forkDaemon(
  refreshLogin.pipe(
    Effect.schedule(Schedule.spaced(refreshInterval))
  )
);
```

See https://effect.website/docs/scheduling/repetition/#skipping-first-execution

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,300 @@
...
+								"PM Insights scheduled login failed; keeping previous token",
+								{
+									message:
+										err instanceof FailedToHandlePmInsightsRequestError
+											? err.message
+											: `${err}`,
+								},
+							);
+						}
+					});
```
Thomasvdam: Maybe I'm missing something, but the way I understand the flow is as follows:

1. Obtain valid token
2. Wait 50m
3. Retry token
  a. Success: store new token
  b. Failure: keep old token
4. Back to step 2

Assuming 1 hour token validity, doesn't this mean if the refresh fails we're stuck with an invalid token for 40 minutes before it retries again?

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -0,0 +1,300 @@
...
+						Effect.mapError(
+							(error) =>
+								new FailedToHandlePmInsightsRequestError({
+									error: `Failed to read response: ${error}`,
+									status: 500,
+								}),
+						),
+					);
+
+					const upstreamOk = response.status >= 200 && response.status < 300;
```
Thomasvdam, line 282: Should we check for auth failures to trigger a refresh of the token?

hacheigriega: Good point.. added

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
Thomasvdam: I think this is still the flow, but now that 400 responses also trigger a token refresh it's less of an issue. 

Maybe something to look at if we notice issues once it's live.

hacheigriega: I added a repeated retry logic (default 5 minute interval) when refresh fails. Let me know what you think.

Thomasvdam: It looks a little funky, but in all fairness I have no idea how to write it in a more elegant way. Probably in a few months when I am better acquainted with Effect.schedule :D

Thomasvdam (approved):

PR #114, opened 2026-05-08
Handle dxFeed events

hacheigriega:
Make use of demo URL to handle events correctly.


workspace/data-proxy/src/modules/dxfeed/schema.ts
```diff
@@ -1,8 +1,57 @@
...
+	}
+
+	const fields = {} as Partial<Record<DxFeedEventField, number>>;
+
+	for (const key of eventFields) {
+		const value = readFiniteNumber(record, key);
+		if (value !== undefined) {
+			fields[key] = value;
+		}
+	}
```
Thomasvdam, line 34: This is not correct since we allow configuring the eventFields in the config. In theory we could only receive a subset of the array we're importing.

hacheigriega: Thanks.. the function now takes event fields from the caller where we pass the runtime config

workspace/data-proxy/src/modules/dxfeed/schema.ts
```diff
@@ -1,8 +1,55 @@
...
+
+	if (fields.bidPrice === undefined && fields.askPrice === undefined) {
+		return undefined;
+	}
```
Thomasvdam: I don't think we should have an opinion on this? If someone only wants to proxy the timestamps that should be allowed, or we should make it clear what that fails

workspace/data-proxy/src/modules/dxfeed/schema.ts
```diff
@@ -1,8 +1,55 @@
...
+
+	const fields = {} as Partial<Record<DxFeedEventField, number>>;
+
+	for (const key of eventFields) {
+		const value = readFiniteNumber(record, key);
+		if (value !== undefined) {
+			fields[key] = value;
+		}
+	}
```
Thomasvdam, line 34: I think proxying transparently is the easiest. Now if we ever extend the eventFields with one that is not a number it won't work and we'll be surprised until we realise we have this check silently discarding any non-numbers :P

Thomasvdam (approved):

PR #116, opened 2026-05-14
feat(data-proxy): add hydromancer module

jaspersagent:
Adds a Hydromancer module with WS subscriptions for `activeAssetCtx` and a REST batch fallback when the cache is stale or the socket is errored. Requests to `POST /info` go through the module: assetContext bodies are served from cache or batch-fetched, anything else returns 400. The branch also lands the Bun-to-Node runtime migration (with a follow-up fix that threads the already-parsed request body through `ModuleService.handleRequest`, since elysia's parse step now consumes the request stream), dxfeed event handling, and a shared idle-cleanup utility used by hydromancer and pyth-lazer.

PR #126, opened 2026-05-21
feat(hydromancer)!: add l2Book surface and body-driven /info

jaspersagent:
`/info` now reads coins from the request body, and the hydromancer module serves order books alongside asset contexts.

The route accepts `{"type": "assetContext" | "l2Book", "coins": [...]}` and rejects anything else with a 400. The proxy parses the body once and threads it into the module handler.

assetContext stays WS-first with a REST fallback. l2Book is WS-first with a short wait for the first frame and no fallback, because books update many times a second and a fallback fetch would be stale before it returned. An unseeded coin comes back `null`. Both surfaces multiplex over one WebSocket on one credential.

The WS code then went through a review pass. It was leaking three ways: book-cache waiters stranded on timeout, reconnect listeners that were never detached, and idle-cleanup fibers with no way to stop them. All three are closed, and the module now runs as a scoped layer so every background daemon dies with the layer. The two near-identical WS channels became one channel-keyed registry, and the `getOrWaitPrice` catch-and-evict block that dxfeed, lo-tech, and pyth-lazer each hand-rolled is now a shared `getOrWaitOrEvict` primitive.

Two things to know before merging:

- The hydromancer config fields were renamed (`coinsCleanup*` to `assetCtx*` and friends), so existing operator config files break.
- `ModuleService.handleRequest` takes a required `body` parameter now; chainlink-streams and pm-insights use it instead of re-reading the request.


jaspersagent:
Closing in favor of #127. This branch had drifted 17 commits behind main, and its first commits duplicated work already merged via #116, so #127 rebuilds the l2Book flow and hardening cleanly on current main.

PR #127, opened 2026-05-21
feat(hydromancer): add l2Book WebSocket order-book surface

jaspersagent:
The hydromancer module serves order books now, alongside asset contexts.

`/info` accepts `{"type": "l2Book", "coins": [...]}` next to the existing `assetContext` shape. l2Book is answered WS-first: the request waits briefly for the first frame and returns `null` for an unseeded coin, with no REST fallback, because books update many times a second and a fallback fetch would be stale before it returned. Both surfaces multiplex over one WebSocket on one credential.

The WS code is built not to leak. One channel-keyed registry carries both channels, and reconnect listeners are detached on close so a reconnect storm cannot strand them. The book cache reuses the shared price cache with a short wait timeout and a `tryGetOrWait` primitive that evicts the waiter once the wait times out.

The second commit is cleanup the shared code makes obvious: dxfeed, lo-tech, and pyth-lazer drop their hand-rolled `getOrWaitPrice` catch-and-evict block for the shared `getOrWaitOrEvict`, and chainlink-streams and pm-insights take the proxy-parsed body instead of re-reading the request. `ModuleService.handleRequest` now requires the body parameter. Config field names are unchanged, so operator `config.json` files keep working.

Verified end to end against Hydromancer mainnet: both `assetContext` and `l2Book` return live data through a running proxy.


workspace/data-proxy/src/modules/hydromancer/hydromancer.test.ts
```diff
@@ -385,35 +410,39 @@ describe("HydromancerModuleService demand-driven subscriptions", () => {
...
 
-		const response = await Effect.runPromise(
-			program.pipe(
-				Effect.provide(HydromancerModuleService(config)),
-				Logger.withMinimumLogLevel(LogLevel.None),
-			),
-		);
-		expect(response.status).toBe(200);
+			yield* Effect.sleep(Duration.millis(0));
+			yield* Effect.sleep(Duration.millis(0));
```
Thomasvdam: Why do we need this? Looks to me like the `flush` we had before also didn't do anything.

workspace/data-proxy/src/modules/hydromancer/hydromancer.test.ts
```diff
@@ -575,13 +603,162 @@ describe("HydromancerModuleService REST fallback when WS is errored", () => {
...
+
+	afterEach(() => {
+		globalThis.WebSocket = originalWebSocket;
+	});
+
+	it("returns the seeded snapshot for a pre-subscribed coin", async () => {
+		const config: HydromancerModuleConfig = {
+			...baseConfig,
+			l2BookSubscriptionCoins: ["BTC"],
+		};
```
Thomasvdam: This doesn't test that it actually sends a subscription message for the initial coins.

workspace/data-proxy/src/utils/idle-cleanup.ts
```diff
@@ -27,23 +31,38 @@ export interface ForkIdleCleanupOptions<K> {
  */
 export const forkIdleCleanup = <K>(
 	options: ForkIdleCleanupOptions<K>,
-): Effect.Effect<Fiber.RuntimeFiber<unknown, unknown>, never, never> =>
-	Effect.forkDaemon(
+): Effect.Effect<Fiber.RuntimeFiber<unknown, unknown>, never, never> => {
+	// Consecutive onExpire failures per key. A key that keeps failing is
+	// force-removed once it reaches the cap so the map cannot grow forever.
```
Thomasvdam: Ok great, now the map won't grow forever but whichever other process relied on this callback to clean up resources is now most likely leaking something that's probably bigger than a key value pair in a map.

Maybe it makes more sense to use this counter to space out retries? So the first X times we attempt it and log a warning, after X we log an error and try every Y (or some capped exponential backoff)

workspace/data-proxy/src/modules/shared/price-cache.ts
```diff
@@ -103,8 +107,28 @@ export const createPriceCache = <K, V>() =>
...
+			);
+
+		// Waits for a value; on timeout or error deletes the cached entry and
+		// rethrows. For callers where a missing value must fail the request.
+		const getOrWaitOrEvict = (key: K) =>
+			getOrWaitPrice(key).pipe(
+				Effect.catchTag("FailedToGetPriceError", (error) =>
+					Effect.zipRight(deletePrice(key), Effect.fail(error)),
+				),
+			);
```
Thomasvdam, line 130: I think this replaces all uses of `getOrWaitPrice`, so we can clean that up.

workspace/data-proxy/src/modules/shared/freshness-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const createFreshnessCache = <K, V>(): Effect.Effect<
+	FreshnessCache<K, V>
+> =>
+	Effect.gen(function* () {
+		const entries = MutableHashMap.empty<K, FreshnessEntry<V>>();
+
+		const set = (key: K, value: V, now: number) =>
+			Effect.sync(() => {
+				MutableHashMap.set(entries, key, { value, lastUpdate: now });
```
Thomasvdam: Is there a reason to make these functions effects?

workspace/data-proxy/src/modules/shared/freshness-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+	lastUpdate: number;
+}
+
+/**
+ * A push-on-write cache whose reads are gated by freshness. `get` returns a
+ * value only when it was written within the supplied staleness window; a
+ * stale or absent key returns `None`, leaving the caller free to fall back
+ * to a pull such as a REST fetch. It has no waiters and no timeouts.
+ */
+export interface FreshnessCache<K, V> {
```
Thomasvdam, line 14: I don't think we're using this interface outside of this file?

workspace/data-proxy/src/modules/hydromancer/ws-client.test.ts
```diff
@@ -120,7 +159,12 @@ class FakeWebSocket extends EventTarget {
 	constructor(url: string) {
 		super();
 		this.url = url;
-		FakeWebSocket.instances.push(this);
+		// Only record sockets this test file opened. A leaked WS daemon from
+		// another test file also constructs FakeWebSockets; filtering by URL
+		// keeps those out of `instances` so the reconnect tests stay reliable.
+		if (url.includes("wsclient.test")) {
+			FakeWebSocket.instances.push(this);
+		}
```
Thomasvdam: I'm not sure how this would happen, given that each test file seems to define their own fake WS class.

workspace/data-proxy/src/modules/hydromancer/ws-client.test.ts
```diff
@@ -155,10 +199,25 @@ class FakeWebSocket extends EventTarget {
...
+	const startedAt = Date.now();
+	while (FakeWebSocket.instances.length < count) {
+		if (Date.now() - startedAt > timeoutMs) {
+			throw new Error(
+				`Timed out waiting for ${count} WebSocket instances; saw ${FakeWebSocket.instances.length}`,
+			);
+		}
+		await new Promise<void>((r) => setTimeout(r, 5));
+	}
+};
```
Thomasvdam, line 220: Can't we test this properly with Effect's TestClock?

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -23,94 +25,127 @@ export const HydromancerModuleService = (config: HydromancerModuleConfig) =>
...
-				route: Route,
-				_params: Record<string, string>,
-				_request: Request,
-				body?: string,
+			const lastRequestToCoin = MutableHashMap.empty<string, number>();
+			const lastRequestToBookCoin = MutableHashMap.empty<string, number>();
+
+			// One entry per WS channel: the demand map the cleanup pass drains,
+			// the cleanup cadence, and the cache eviction that runs before the
+			// channel is unsubscribed.
```
Thomasvdam: My reading comprehension is not sufficient enough to get what this means :P

workspace/data-proxy/src/modules/hydromancer/hydromancer.test.ts
jaspersagent: Right that the old `flush()` was a no-op for tests that only awaited the runPromise result. These sleeps weren't, because we now observe `FakeWebSocket.instances[0]` from *inside* the program: the daemon owns the socket and the layer scope tears it down on exit, so the forkDaemon fiber has to be scheduled before the first `new WebSocket(...)` is visible.

Replaced with a `waitForSocket` polling helper + `Effect.yieldNow()`, so the reason for each wait is explicit.

workspace/data-proxy/src/modules/hydromancer/hydromancer.test.ts
jaspersagent: Good catch. Added an explicit `expect(sentOnOpen).toEqual([buildSubscribeFrame("l2Book", "BTC")])` before the response check; the test now fails if the subscribe pipeline becomes a no-op.

workspace/data-proxy/src/utils/idle-cleanup.ts
jaspersagent: Agreed. Reworked: a failing `onExpire` keeps its entry forever and the counter only escalates log severity (warn for the first 3, error after). Retries happen on every cleanup pass at the configured interval, so a transient downstream failure resolves itself. Added a test that drives 5 failing passes and asserts the entry persists.

workspace/data-proxy/src/modules/shared/price-cache.ts
jaspersagent: Done. `getOrWaitPrice` is a closure-local helper now, not exported. Rewrote `price-cache.test.ts` against the wrappers (`tryGetOrWait`, `getOrWaitOrEvict`); the waiter mechanism is still fully covered via the public surface.

workspace/data-proxy/src/modules/shared/freshness-cache.ts
jaspersagent: No good reason. They were all `Effect.sync`. Dropped the wrappers: `set/get/remove` are plain sync functions now, callers stopped yielding on them. The factory still returns `Effect.Effect<FreshnessCache<K,V>>` so tests construct it the same way.

workspace/data-proxy/src/modules/shared/freshness-cache.ts
jaspersagent: It is. `ws-client.ts:127` declares `createHydromancerWS(config, assetCache: FreshnessCache<string, AssetCtx>, ...)`, so the type import is genuinely consumed. Keeping it exported.

workspace/data-proxy/src/modules/hydromancer/ws-client.test.ts
jaspersagent: Each file has its own `FakeWebSocket` class, but `createHydromancerWS` forks its connect loop via `Effect.forkDaemon`, which is intentionally detached from the layer scope. A daemon started by another file outlives its `runPromise`; when bun switches to this file, the daemon's next reconnect calls `new globalThis.WebSocket(...)`, which is now THIS file's class. The URL filter discriminates by the leaked daemon's config-time `wsUrl` (captured at fiber start). Expanded the inline comment to spell this out.

workspace/data-proxy/src/modules/hydromancer/ws-client.test.ts
jaspersagent: Yes in principle. The reconnect `Schedule.spaced` runs on Effect's clock, so `TestClock.adjust` would advance it. The friction is the FakeWebSocket↔daemon bridge: `triggerOpen`/`triggerClose` dispatch synchronously and the daemon's listeners hop back via `Runtime.runSync`, so we'd still need yields between the external event and the Effect-driven reaction. Leaving the URL-keyed polling for this PR; happy to do the TestClock refactor as a follow-up.

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
jaspersagent: Rewritten. Now names the locals each entry carries (`lastRequest`, `ttl/interval`, `onEvict`) and what the loop below uses them for.

mariocao (commented):
@jasperdg , what's the status of this PR?
Do we still want to merge it?

hacheigriega:
Superseded by #171

PR #129, opened 2026-05-25
feat(pyth): add real time channel to config

mariocao:
Self-explanatory

### DIFF
--- a/workspace/data-proxy/package.json
+++ b/workspace/data-proxy/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@seda-protocol/data-proxy",
 	"main": "./src/index.ts",
-	"version": "2.7.3",
+	"version": "2.7.4",
 	"devDependencies": {
 		"@types/big.js": "^6.2.2",
 		"@types/ws": "^8.18.1",
--- a/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
+++ b/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
@@ -11,7 +11,12 @@ export const PythLazerModuleConfigSchema = v.strictObject({
 		}),
 	),
 	channel: v.optional(
-		v.picklist(["fixed_rate@200ms", "fixed_rate@50ms", "fixed_rate@1000ms"]),
+		v.picklist([
+			"real_time",
+			"fixed_rate@50ms",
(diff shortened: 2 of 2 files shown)


gluax (commented):
Shouldn't we make this done at the asset level?

Otherwise the config is set to a channel that other assets don't possibly have? Or ig the intent is to have different route for the different channels?

mariocao:
> Shouldn't we make this done at the asset level?
> 
> Otherwise the config is set to a channel that other assets don't possibly have? Or ig the intent is to have different route for the different channels?

I don't fully get your comment.

This data proxy module subscribes via WS to assets for a given channel, that's why I believe it needs to be at module level, and not asset level (like with a param).

This means that we would expose 1 endpoint for each channel supported.
If I missed something obvious, please tell me.

gluax:
> > Shouldn't we make this done at the asset level?
> 
> > 
> 
> > Otherwise the config is set to a channel that other assets don't possibly have? Or ig the intent is to have different route for the different channels?
> 
> 
> 
> I don't fully get your comment.
> 
> 
> 
> This data proxy module subscribes via WS to assets for a given channel, that's why I believe it needs to be at module level, and not asset level (like with a param).
> 
> 
> 
> This means that we would expose 1 endpoint for each channel supported.
> 
> If I missed something obvious, please tell me.

Ah I was trying to avoid more computation on the OP side cause if we just forward what channel we should be subscribed to it's less looping to separate different assets with different channels.

mariocao:
> > > Shouldn't we make this done at the asset level?
> > 
> > 
> > > 
> > 
> > 
> > > Otherwise the config is set to a channel that other assets don't possibly have? Or ig the intent is to have different route for the different channels?
> > 
> > 
> > I don't fully get your comment.
> > This data proxy module subscribes via WS to assets for a given channel, that's why I believe it needs to be at module level, and not asset level (like with a param).
> > This means that we would expose 1 endpoint for each channel supported.
> > If I missed something obvious, please tell me.
> 
> Ah I was trying to avoid more computation on the OP side cause if we just forward what channel we should be subscribed to it's less looping to separate different assets with different channels.

In the OP side would be a different config URL path for pythpro, no new logic.

gluax (approved):

gluax:
That's way simpler than what I was thinking aha. Sorry. LGTM

PR #130, opened 2026-05-26
chore: always give back prices even when one symbol fails

FranklinWaller:
## Motivation

Sometimes a price could not be fetched and the whole request would fail. Now we flag the failing price and give back the other prices. This way one bad price doesn't stall the rest of the prices.

## Explanation of Changes

Add `__sedaHasPrice: boolean` to the response so the OP can detect if there is a price available.

### DIFF
--- a/workspace/data-proxy/package.json
+++ b/workspace/data-proxy/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@seda-protocol/data-proxy",
 	"main": "./src/index.ts",
-	"version": "2.7.4",
+	"version": "2.7.5",
 	"devDependencies": {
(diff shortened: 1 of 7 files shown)


Thomasvdam (approved):
Alternatively we could do an envelop, but that'd be breaking so I understand why we went this route.

The change is not required for hydro?

FranklinWaller:
> Alternatively we could do an envelop, but that'd be breaking so I understand why we went this route.
> 
> The change is not required for hydro?

Yes, I considered it, but this seemed the most normal route.

Hydro has already a null system (it's completely different from the rest..)

PR #131, opened 2026-05-26
LO:TECH response involving failed symbols

hacheigriega:
## Explanation of Changes

Explanation of the three commits in order:
1. Apply changes from #130 to the lo-tech module. Also, move cleanup upon timeout error from the call sites to the shared cache.
2. Lo-tech module now handles acknowledgement messages. The price feed map is now updated only after receiving ack messages, preventing updates when a subscription request fails.
3. Lo-tech module also handles subscription failure errors. This helps us clean up the price feed IDs we add for every subscription request.
4. Price fetch is now done concurrently so that timeouts do not accumulate across multiple invalid symbols.

## Testing

```bash
$ curl http://localhost:5384/proxy/lotech/NVDA,WTIN6 | jq

[
  {
    "type": "PRICE",
    "symbol": "NVDA",
    "ingress_ts": 1779818875945310,
    "publish_ts": null,
    "transaction_ts": 1779818875860000,
    "price": 213.496,
    "spread": 0.01,
    "__sedaHasPrice": true
  },
  {
    "symbol": "WTIN6",
    "__sedaHasPrice": false
  }
]
```

### DIFF
--- a/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
+++ b/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
@@ -24,9 +24,11 @@ import { FailedToHandleLoTechRequestError } from "./errors";
 import type {
 	LoTechAck,
 	LoTechDataPrice,
+	LoTechErrorMessage,
 	LoTechParsedData,
 	LoTechResponse,
 } from "./schema";
+import { LoTechSubscriptionFailureCode } from "./schema";
 import { makeLoTechWebSocketService } from "./ws-client";
 
 export type PriceFeedSymbol = string;
@@ -94,6 +96,19 @@ export const LoTechModuleService = (config: LoTechModuleConfig) =>
 					MutableHashMap.set(lastRequestToPriceFeed, symbol.value, now);
 				});
 
+			const handleErrorMessage = (msg: LoTechErrorMessage) =>
+				Effect.gen(function* () {
+					if (msg.error.code === LoTechSubscriptionFailureCode) {
+						yield* Effect.logInfo("Subscription request failed", {
+							id: msg.error.id,
+						});
+						MutableHashMap.remove(priceFeedIds, msg.error.id);
+						return;
+					}
+
+					yield* Effect.logWarning("Unexpected LO:TECH error message", { msg });
+				});
+
 			const loTechWs = yield* makeLoTechWebSocketService({
 				config,
 				runtime,
@@ -116,6 +131,7 @@ export const LoTechModuleService = (config: LoTechModuleConfig) =>
 					}),
 				handleDataMessage,
 				handleAckMessage,
+				handleErrorMessage,
 			});
 
 			const start = () =>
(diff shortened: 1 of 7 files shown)


workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -64,6 +76,24 @@ export const LoTechModuleService = (config: LoTechModuleConfig) =>
...
+						yield* Effect.logWarning("Received ack for unknown price feed ID", {
+							msg,
+						});
+						return;
+					}
+
+					MutableHashMap.set(priceFeeds, symbol.value, msg.ack.id);
+
+					const now = yield* Clock.currentTimeMillis;
+					MutableHashMap.set(lastRequestToPriceFeed, symbol.value, now);
```
Thomasvdam, line 96: Is this correct? It reads a little weird that a message from lo-tech->proxy updates a map that tracks requests from client->proxy

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -31,9 +38,16 @@ export const LoTechModuleService = (config: LoTechModuleConfig) =>
 			yield* Effect.logInfo("Initializing LO:TECH module");
 
 			const runtime = yield* Effect.runtime();
+
+			// Monotonically increasing counter for price feed IDs.
+			let latestPriceFeedId = 0;
+			// Map from price feed IDs to symbols. (Used to process acks)
+			const priceFeedIds = MutableHashMap.empty<number, PriceFeedSymbol>();
```
Thomasvdam, line 47: What are acks for? 😬

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: So with this change we update the price feed subscription map and the last request map upon receiving ack to avoid updating these maps for invalid symbols that we have failed to subscribe for. Previously, we were tracking last request times for all the invalid symbols we receive from clients, even though there were no active price feeds for those.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: I think the previous comment addresses this, but please let me know what you think.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: Ah check, so if I understand it correctly the flow now is:

- Request for symbol `X`
- If new symbol, create new feedId (monotonically increasing counter) and store in `priceFeedIds`
- Receive ack for an ID
- Check if `priceFeedIds` contains the ID
- If yes, make the pricefeed 'official' by adding mapping entry in `priceFeeds` 
- And mark it as the first request by setting a timestamp in `lastRequestToPriceFeed`

If this is correct I find it a little hard to get this from the code, but I also don't have an idea on how to express it differently.

I do have one question; what happens to `priceFeedIds` that never receive an ack? Is that a potential source of an unbounded map that doesn't get cleaned?

hacheigriega: Yes and yes, unfortunately. The ack message doesn't contain the symbol, so there is no way to avoid storing a symbol in `priceFeedIds` every time we send a subscription request. 

But now that I think about it, we could handle this issue by handling the subscription failure errors we receive from the upstream. I will try that

Thomasvdam (approved):

PR #137, opened 2026-05-29
Add pyth lazer latest price surface

jaspersagent:
## Summary

- Add the POST `/v1/latest_price` pyth-lazer surface expected by deployed pyth_pro oracle programs.
- Keep the existing GET `/price/:symbols` array surface backed by the same websocket cache.
- Preserve all-or-fail request ordering for latest_price so response entries align with requested feed positions.
- Store per-feed stream timestamps in cache so `parsed.timestampUs` is derived from the returned feeds, not unrelated updates.

## Validation

- `bun test workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.test.ts`
- `bunx biome check workspace/data-proxy/src/config/pyth-lazer-module-config.ts workspace/data-proxy/src/controllers/proxy/handle-proxy-request.ts workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.test.ts`
- `bun run check-ts`
- `bun test`

### DIFF
--- a/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
+++ b/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
@@ -50,7 +50,10 @@ export interface PythLazerModuleConfig
 export const PythLazerModuleRouteSchema = v.strictObject({
 	...RouteSchema.entries,
 	moduleName: v.string(),
-	fetchFromModule: v.string(),
+	// Present: the symbols/ids come from the path (e.g. GET /price/:symbols),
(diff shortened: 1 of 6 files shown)


hacheigriega (commented):
Looks good to me. I did wonder if it makes sense to have `handlePathRequest` be a part of the module but not `handleLegacyLatestPriceRequest`. But ultimately I didn't see a good reason against it.

hacheigriega:
Will open a new PR with an updated implementation. 
What was the motivation for this implementation?

PR #139, opened 2026-06-08
feat(binance): add Binance websocket price-relay module

jaspersagent:
Adds a Binance market-data module that relays websocket price frames through the proxy, following the same shape as the existing hydromancer and lo-tech stream modules: a config schema with a no-op route validator, a reconnecting ws-client, and a service backed by the shared price-cache and idle-cleanup. A request subscribes any not-yet-tracked symbols, waits on the shared cache, and returns one entry per requested symbol in request order, each flagged with `__sedaHasPrice` so callers reading by index stay aligned. Symbols nobody has requested within the TTL are unsubscribed and evicted by the shared idle-cleanup daemon. Stream type (bookTicker, aggTrade, trade, ticker, miniTicker) and reconnect backoff are config-driven, and subscribe frames are batched one-per-call to stay under Binance's 5/sec control-message cap.

- Binance public market-data streams need no credentials, so the config parser's match arm adds no env secrets, unlike the credentialed modules.
- The ws-client's connect/reconnect machinery is a third near-copy of the hydromancer/lo-tech pattern. Extracting a shared `reconnecting-ws` helper is deliberately left to a follow-up rather than bundled here.

### DIFF
--- a/workspace/data-proxy/src/modules/binance/binance.ts
+++ b/workspace/data-proxy/src/modules/binance/binance.ts
@@ -90,6 +90,9 @@ export const BinanceModuleService = (config: BinanceModuleConfig) =>
 					}
 
 					const now = yield* Clock.currentTimeMillis;
+					// When the socket has errored its cached prices may be stale, so
+					// do not vouch for them (mirrors the hydromancer module).
+					const socketHealthy = !(yield* ws.hasError());
 					const newSymbols: string[] = [];
 					for (const requested of requestedSymbols) {
 						const symbol = requested.toUpperCase();
@@ -116,7 +119,7 @@ export const BinanceModuleService = (config: BinanceModuleConfig) =>
 						const requested = requestedSymbols[i];
 						const result = results[i];
 
-						if (Either.isLeft(result)) {
+						if (Either.isLeft(result) || !socketHealthy) {
 							prices.push({ symbol: requested, [HAS_PRICE_KEY]: false });
 						} else {
 							prices.push({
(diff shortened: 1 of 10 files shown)


hacheigriega (approved):
LGTM but I haven't been run it myself because it kept disconnecting:
```
2026-06-15T18:30:52.726Z WARN  #17 Binance WS disconnected
{
  "reason": "close"
}
```
Maybe the endpoint is blocked due to my location?

workspace/data-proxy/src/modules/binance/binance.ts
```diff
@@ -0,0 +1,146 @@
...
+					if (requestedSymbols.length > config.maxSymbolsPerRequest) {
+						return yield* Effect.fail(
+							new FailedToHandleBinanceRequestError({
+								error: `Too many symbols, max is ${config.maxSymbolsPerRequest} but got ${requestedSymbols.length}`,
+								status: 400,
+							}),
+						);
+					}
+
+					const now = yield* Clock.currentTimeMillis;
```
mariocao, line 92: Just double-checking: some modules add this check and others not.
Which one is best?

```suggestion
					const now = yield* Clock.currentTimeMillis;
					const socketHealthy = !(yield* ws.hasError());
```

jaspersagent: Good catch on the inconsistency. The hydromancer pattern is the one to follow: treat cached prices as not-live when the socket has errored. Done in ffd8922, compute `socketHealthy = !(yield* ws.hasError())` and flag a cached symbol as `__sedaHasPrice: false` while the socket is down, so a frozen last price is never served during a disconnect. Added a test that drops the socket and asserts a previously-cached symbol comes back as a miss.

PR #140, opened 2026-06-09
feat(lighter): add Lighter websocket ticker-relay module

jaspersagent:
Add a price-relay module for Lighter's perp DEX, mirroring the existing websocket stream modules. Requests carry Lighter's numeric market ids directly (1 for BTC, 110 for NVDA); the module subscribes to `ticker/{id}` over one reconnecting websocket, caches each frame by market id, and returns one entry per requested id flagged with `__sedaHasPrice`. A proactive ping holds the connection under Lighter's two-minute idle timeout, and markets nobody is requesting are dropped by the shared idle-cleanup daemon. It relays public market data, so the module needs no credentials.

- The request contract is market-id-based (`/price/1,110`), unlike the symbol-based sibling modules. Callers map a symbol to its id out of band via Lighter's `/api/v1/orderBooks`.
- A non-integer token short-circuits to a miss; a valid-but-nonexistent numeric id just yields no price (Lighter replies with a harmless error frame and the socket stays up).

### DIFF
--- a/workspace/data-proxy/src/modules/lighter/lighter.ts
+++ b/workspace/data-proxy/src/modules/lighter/lighter.ts
@@ -98,6 +98,9 @@ export const LighterModuleService = (config: LighterModuleConfig) =>
 					}));
 
 					const now = yield* Clock.currentTimeMillis;
+					// When the socket has errored its cached prices may be stale, so
+					// do not vouch for them (mirrors the hydromancer module).
+					const socketHealthy = !(yield* ws.hasError());
 					const newMarketIds: number[] = [];
 					for (const { marketId } of requested) {
 						if (marketId === null) continue;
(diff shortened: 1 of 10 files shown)


hacheigriega (approved):

workspace/data-proxy/src/modules/lighter/lighter.ts
```diff
@@ -0,0 +1,166 @@
...
+							}),
+						);
+					}
+
+					const requested = requestedTokens.map((token) => ({
+						token,
+						marketId: parseMarketId(token),
+					}));
+
+					const now = yield* Clock.currentTimeMillis;
```
mariocao, line 100: We could also add the `ws.hasError` check in here.

jaspersagent: Added in 16eefe4. The Lighter WS client had no `hasError()` yet, so I added one (true while there is no open connection, since `currentWS` is nulled on every disconnect and set on open) and gated cached prices on it in the handler, matching the hydromancer and binance modules. Test included: drop the socket, and a previously-cached market comes back as a miss instead of a stale price.

PR #142, opened 2026-06-11
LO:TECH support for multiple exchanges

hacheigriega:
## Motivation

To support futures in addition to US equities, we need the LO:TECH module to establish and maintain multiple WebSocket connections.

## Explanation of Changes

To prevent unbounded creation of WebSocket connections, we use configuration to specify a list of supported exchanges. By default, it is set to `futures` and `us_equities`.

## Testing

Example configuration:
```json
{
	"modules": [
		{
			"type": "lo-tech",
			"baseUrl": "wss://data.lo.tech/ws/v1",
			"name": "lotech",
			"supportedExchanges": ["us_equities", "futures"],
			"loTechApiKeyEnvKey": "LOTECH_API_KEY",
			"priceFeeds": [
				{
					"symbol": "NVDA",
					"exchange": "us_equities",
					"dataType": "PRICE"
				},
				{
					"symbol": "WTIN6",
					"exchange": "futures",
					"dataType": "PRICE"
				}
			]
		}
	],
	"routes": [
		{
			"type": "lo-tech",
			"moduleName": "lotech",
			"path": "/:exchange/:priceSymbol",
			"method": "GET",
			"fetchFromModule": "{:priceSymbol}"
		}
	]
}

```

Example requests:
```bash
curl http://localhost:5384/proxy/us_equities /NVDA
curl http://localhost:5384/proxy/futures/WTIN6
```

### DIFF
--- a/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
+++ b/workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
@@ -13,7 +13,9 @@ import {
 import type { Route } from "../../config/config-parser";
 import {
 	LO_TECH_DATA_TYPE_PRICE,
+	LO_TECH_EXCHANGE_PATH_PARAM,
 	type LoTechModuleConfig,
+	assertSupportedLoTechExchange,
 } from "../../config/lo-tech-module-config";
 import { HAS_PRICE_KEY } from "../../constants";
 import { createErrorResponse } from "../../controllers/create-error-response";
@@ -31,7 +33,51 @@ import type {
 import { LoTechSubscriptionFailureCode } from "./schema";
 import { makeLoTechWebSocketService } from "./ws-client";
 
-export type PriceFeedSymbol = string;
+export type PriceFeedKey = `${string}:${string}`;
+
+type PriceFeedSubscription = {
+	symbol: string;
+	exchange: string;
+};
+
+export const priceFeedKey = (exchange: string, symbol: string): PriceFeedKey =>
+	`${exchange}:${symbol}`;
+
+export const parsePriceFeedKey = (key: PriceFeedKey): PriceFeedSubscription => {
+	const separatorIndex = key.indexOf(":");
+	return {
+		exchange: key.slice(0, separatorIndex),
+		symbol: key.slice(separatorIndex + 1),
+	};
+};
+
+// Resolve the exchange for the given request from the required path parameter.
+export const resolveLoTechExchange = (
+	exchange: string | undefined,
+	moduleConfig: Pick<LoTechModuleConfig, "supportedExchanges">,
+): Effect.Effect<string, FailedToHandleLoTechRequestError> =>
+	Effect.gen(function* () {
+		if (exchange === undefined || exchange === "") {
+			return yield* Effect.fail(
+				new FailedToHandleLoTechRequestError({
+					error: `Missing required "${LO_TECH_EXCHANGE_PATH_PARAM}" path parameter`,
+				}),
+			);
+		}
+
+		yield* assertSupportedLoTechExchange(
+			exchange,
+			moduleConfig.supportedExchanges,
+		).pipe(
+			Effect.mapError(
+				(error) =>
+					new FailedToHandleLoTechRequestError({
+						error,
+					}),
+			),
+		);
+		return exchange;
+	});
 
 export const LoTechModuleService = (config: LoTechModuleConfig) =>
 	Layer.effect(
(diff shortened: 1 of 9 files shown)


workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -31,7 +33,72 @@ import type {
 import { LoTechSubscriptionFailureCode } from "./schema";
 import { makeLoTechWebSocketService } from "./ws-client";
 
-export type PriceFeedSymbol = string;
+export type PriceFeedKey = string;
```
Thomasvdam: It's not perfect and doesn't impact runtime, but you can type this as 
```ts
type PriceFeedKey = `${string}:${string}`;
```

Indicates a little better what the intention is without runtime overhead.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -31,7 +33,72 @@ import type {
...
+
+export const priceFeedKey = (exchange: string, symbol: string): PriceFeedKey =>
+	`${exchange}:${symbol}`;
+
+export const parsePriceFeedKey = (key: PriceFeedKey): PriceFeedSubscription => {
+	const separatorIndex = key.indexOf(":");
+	return {
+		exchange: key.slice(0, separatorIndex),
+		symbol: key.slice(separatorIndex + 1),
+	};
```
Thomasvdam, line 51: Not sure if there's a (meaningful) performance difference but you can do:

```typescript
const [exchange, symbol] = key.split(":");
```

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -31,7 +33,72 @@ import type {
...
+			(priceFeed) => priceFeed.symbol === symbol,
+		);
+		if (configuredFeed !== undefined) {
+			return configuredFeed.exchange;
+		}
+
+		if (moduleConfig.defaultExchange === undefined) {
+			return yield* Effect.fail(
+				new FailedToHandleLoTechRequestError({
+					error: `No exchange for symbol "${symbol}". Set "defaultExchange", add the symbol to "priceFeeds", or pass the "${LO_TECH_EXCHANGE_QUERY_PARAM}" query parameter`,
```
Thomasvdam: It feels odd to tell a calling client that they should update the config of the proxy.

We should either make the query parameter required or make it so the module always has a default exchange. We could do something like:

```
1. Use LO_TECH_EXCHANGE_QUERY_PARAM if present
2. Use defaultExchange if set
3. Use first entry in supported exchanges
```

But thinking about it I feel like requiring the client to specify the exchange is the easiest to reason about for both the client and the proxy operator.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
```diff
@@ -275,6 +416,11 @@ export const LoTechModuleService = (config: LoTechModuleConfig) =>
 					);
 				}).pipe(
 					Effect.withSpan("handleLoTechRequest"),
+					// Effect.catchIf(
+					// 	(error): error is FailedToHandleLoTechRequestError =>
+					// 		error instanceof FailedToHandleLoTechRequestError,
+					// 	(error) => Effect.succeed(createErrorResponse(error, error.status)),
+					// ),
```
Thomasvdam: Should this be removed or put back?

hacheigriega: Oops, totally forgot. I put it back.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Actually I think indexOf makes more sense here for keys involving multiple colons like `futures:BTC-USDT:SPOT`. I'm not sure if such symbols exist, but just in case?

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Yeah agreed. We now require exchange for every symbol in the config or request. Also removed defaultExchange from config.

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: Great catch! I didn't consider that :D

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
Thomasvdam: There's an argument to be made for changing it to a path parameter rather than a query parameter; `/proxy/<exchange>/<symbol>` for example.

But I don't think it's that big a deal, there are plenty of APIs that have required query params these days.

Thomasvdam (approved):

workspace/data-proxy/src/modules/lo-tech/lo-tech.ts
hacheigriega: Ah yeah, originally chose query parameter because it was optional, but path parameter makes more sense now. Updated.

PR #143, opened 2026-06-15
feat(pyth-lazer): coalesce feeds into one bulk subscription

jaspersagent:
Per-feed Pyth Lazer subscriptions make the upstream send one WS frame per feed per channel tick, which saturates the single event loop: request latency tails and price staleness both balloon, and the congestion spills onto unrelated routes that share the proxy process.

### DIFF
--- a/workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
+++ b/workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
@@ -10,7 +10,6 @@ import {
 import {
 	Clock,
 	Data,
-	Duration,
 	Effect,
 	Either,
 	Layer,
@@ -42,10 +41,7 @@ import { getPriceIdBySymbol } from "./get-symbol-price-id";
 // upstream then sends one frame per channel tick carrying all feeds instead of
 // one frame per feed, which keeps the standing WebSocket load (and the price
 // staleness caused by thousands of per-feed frames) flat as feeds grow.
-const BULK_COMPACT_MS = Number(
-	process.env.PYTH_LAZER_BULK_COMPACT_MS ?? 60_000,
-);
-const BULK_OVERLAP_MS = Number(process.env.PYTH_LAZER_BULK_OVERLAP_MS ?? 1_000);
+// Cadence is configured per module via bulkCompactInterval and bulkOverlap.
 
 export class FailedToCreateLazerClientError extends Data.TaggedError(
 	"FailedToCreateLazerClientError",
@@ -55,6 +51,7 @@ export class FailedToCreateLazerClientError extends Data.TaggedError(
 
 type PriceFeedId = number;
 type PriceFeedSymbol = string;
+type SubscriptionId = number;
 
 interface PriceFeedWithSymbol extends ParsedFeedPayload {
 	symbol?: string;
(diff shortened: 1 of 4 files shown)


workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -85,230 +88,397 @@ export const PythLazerModuleService = (config: PythLazerModuleConfig) =>
...
+			);
 
-							if (MutableHashMap.has(subscriptions, newPriceFeedId)) {
-								yield* Effect.logDebug(
-									`Price feed ${newPriceFeedId} is already subscribed`,
-								);
-								return;
-							}
+			sendSubscribe(newBulkSubscriptionId, [...newFeedIds]);
+			yield* Effect.sleep(config.bulkOverlap);
```
mariocao, line 309: Not sure a `sleep` is the best approach here.

workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -85,230 +88,397 @@ export const PythLazerModuleService = (config: PythLazerModuleConfig) =>
...
+				if (Option.isNone(current)) {
+					continue;
+				}
+				// A feed re-subscribed during the overlap holds a fresh individual
+				// subscription that the bulk now supersedes; drop it so it is not
+				// orphaned (no feed would map to it, so no later pass could).
+				if (
+					current.value !== newBulkSubscriptionId &&
+					individualSubscriptionIds.has(current.value)
+				) {
```
hacheigriega, line 331: Woundn't the first condition here always hold since we never assign `newBulkSubscriptionId` to the feeds during compaction?

workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -85,230 +88,397 @@ export const PythLazerModuleService = (config: PythLazerModuleConfig) =>
...
+					deliveredFeedIds.has(feedId)
+				) {
+					absorbable.set(feedId, subId);
+				}
+			}
+
+			const carriedOver = new Set<PriceFeedId>();
+			let droppedFromBulk = 0;
+			for (const feedId of bulkFeedIds) {
+				if (MutableHashMap.has(subscriptions, feedId)) {
```
hacheigriega, line 278: Technically, this check might not enough because a feed that's been cleaned up due to idleness could be re-subscribed. Under this scenario, we would be carrying over the feed without checking `deliveredFeedIds`.

hacheigriega: I suppose we could just assume that a feed that has been included in a bulk has been delivered at least once. Or I think we can just add a subscription ID check?
```typescript
for (const feedId of bulkFeedIds) {
	const subId = MutableHashMap.get(subscriptions, feedId);
	if (Option.isNone(subId)) {
		// Feed has been cleaned up due to idleness
		droppedFromBulk++;
	} else if (subId.value === bulkSubscriptionId) {
		// Carry over
		carriedOver.add(feedId);
	} else {
		// Feed is back on an individual subscription through re-subscribe
		// Do not carry over and let it be handled as an absorbable case.
		droppedFromBulk++;
	}
}
```

hacheigriega:
Will open an updated version of this in #165

PR #144, opened 2026-06-23
feat(multi): add a multi route that fans out to several modules

jaspersagent:
Adds a generic `multi` route that forwards one inbound request to a configured list of named sub-fetches and returns each module's raw response side by side, so a single endpoint can pull from several modules (e.g. binance, lighter, hydromancer) at once. Each fetch targets any module by name with its own `fetchFromModule` path template and/or `body` template, filled from the inbound request via `replaceParams`; the proxy reuses each target module's own handler, so it stays type-agnostic and does no interpretation or blending. Fetches run concurrently, a per-fetch failure or miss becomes a non-fatal `{error, status}` entry rather than failing the request, and multi is exempt from the `__sedaHasPrice` gate since per-source misses are expected. The combined object is signed once like any other route.

Example: `GET /multi/:symbol/:market` with fetches `binance` (`{:symbol}USDT`) and `lighter` (`{:market}`) returns `{ "binance": {...}, "lighter": {...} }`. Startup validation rejects a fetch that references an unknown module, mismatches the module's type, or uses a template variable the path does not provide. Verified with unit tests for the fan-out and config validation, and a live smoke test (happy path, partial miss returning 200, and signature headers present).

### DIFF
--- a/workspace/data-proxy/src/config/config-parser.ts
+++ b/workspace/data-proxy/src/config/config-parser.ts
@@ -45,6 +45,10 @@ import {
 } from "./lo-tech-module-config";
 import { type Modules, ModulesSchema } from "./module-config";
 import {
+	type MultiModuleRoute,
+	MultiModuleRouteSchema,
+} from "./multi-module-config";
+import {
 	type PmInsightsModuleRoute,
 	PmInsightsModuleRouteSchema,
 	validatePmInsightsModuleRoute,
@@ -142,6 +146,7 @@ const ConfigSchema = v.strictObject(
 				PmInsightsModuleRouteSchema,
 				BinanceModuleRouteSchema,
 				LighterModuleRouteSchema,
+				MultiModuleRouteSchema,
 			]),
 		),
 		baseURL: maybe(v.string()),
@@ -179,7 +184,8 @@ export type Route =
 	| LoTechModuleRoute
 	| PmInsightsModuleRoute
 	| BinanceModuleRoute
-	| LighterModuleRoute;
+	| LighterModuleRoute
+	| MultiModuleRoute;
 
 export interface Config extends v.InferOutput<typeof ConfigSchema> {
 	modules: Modules[];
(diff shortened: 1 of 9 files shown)


hacheigriega (approved):

workspace/data-proxy/src/config/config-parser.ts
```diff
@@ -306,6 +312,45 @@ export const parseConfig = (
...
+					}
+					if (target.type !== fetch.type) {
+						return [
+							Result.err(
+								`Multi route fetch "${fetch.name}" has type "${fetch.type}" but module "${fetch.moduleName}" is "${target.type}"`,
+							),
+							hasWarnings,
+						];
+					}
+					// Every {:var} used by a fetch template must be provided by the path.
```
hacheigriega, line 348: I notice that it cannot catch params that a target module reads directly from params without a corresponding template — like lo-tech's `exchange`. Such issues would only be caught at runtime at request time.  I can refactor configs structure later.

workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		const combined: Record<string, unknown> = {};
+		for (const [name, value] of entries) {
+			combined[name] = value;
+		}
+
+		return new Response(JSON.stringify(combined), {
+			status: 200,
+			headers: { "Content-Type": "application/json" },
+		});
```
mariocao, line 98: Just double checking, it returns 200 OK, even if all fetches fail, right?

workspace/data-proxy/src/config/config-parser.ts
```diff
@@ -306,6 +312,45 @@ export const parseConfig = (
 				continue;
 			}
 
+			if (route.type === "multi") {
+				for (const fetch of route.fetches) {
```
mariocao, line 317: Small thing: if two fetches use the same name, the later one overwrite the first in combined response.
See handle-multi-request line 90-92.

Maybe worth to validate unique names at parse time?

workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
```diff
@@ -0,0 +1,99 @@
...
+						type: fetch.type,
+						moduleName: fetch.moduleName,
+						fetchFromModule,
+						path: route.path,
+						method: route.method,
+						allowedQueryParams: fetch.allowedQueryParams ?? [],
+						headers: {},
+						useLegacyJsonPath: true,
+						forwardResponseHeaders: new Set<string>(),
+					} as unknown as Route;
```
mariocao: I believe we don't need the `as unknown`
```suggestion
					} as Route;
```

mariocao (approved):
I can approve, but ideally check my comments first.

workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
jaspersagent: Yes, intentional. `multi` is a fan-out aggregator and is exempt from the `__sedaHasPrice` gate, so each sub-fetch returns its own `{error, status}` entry and the caller reads results per source. A per-source miss is expected, so the combined object is signed and returned with 200 even if every fetch errored. Flipping to non-200 only when all fail would be an arbitrary threshold and inconsistent with the partial-miss case, so I kept it uniform. If you would rather surface an all-failed signal I can add one, but I lean toward leaving it.

workspace/data-proxy/src/config/config-parser.ts
jaspersagent: Good catch. Fixed in 614c05a: the parser now rejects duplicate fetch names in a multi route at parse time, with a test.

workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
jaspersagent: Done in 7f4ade5. `as Route` typechecks, so the `as unknown` is gone.

mariocao (approved):
lgtm

PR #146, opened 2026-07-02
Support Fast proof without chain ID

### LINKED ISSUE #145: 🔧 Support both testnet and mainnet
## 🔧 Chore

Upon checking the codebase, I saw that Fast cannot support both testnet and mainnet under a single process because client signatures are checked against the configured network. 

We can change this behavior by refactoring `decodeSedaFastProof` to perform chain ID check only if a specific network has been configured.

hacheigriega:
## Explanation of Changes

Support decoding Fast proof with or without chain ID to accommodate [this change](https://github.com/sedaprotocol/seda-fast/pull/414). This mode is only supported under `fastOnly` mode since core support requires a specific RPC endpoint to be configured.

|   | DP with chain ID configured | DP without chain ID configured |
| ------------- | ------------- | ------------- |
| Fast proof with chain ID  | Chain ID is checked  | Not checked  |
| Fast proof without chain ID  | Not checked  | Not checked  |

I was thinking that we can have this structure until chain ID-based proof is fully removed from all Fast processes in production. Then, we can remove chain ID-based proof check from data proxy as well.

## Testing

Added unit tests using test vectors created from Fast repo.

## Related PRs and Issues

Closes: #145

### DIFF
--- a/workspace/data-proxy-sdk/src/services/decode-seda-fast-proof.ts
+++ b/workspace/data-proxy-sdk/src/services/decode-seda-fast-proof.ts
@@ -9,7 +9,7 @@ import { FailedToDecodeSedaFastProofError } from "../errors";
 // "{unixTimestampMs}:{signatureAsHexString}" — signed over keccak256(timestamp)
 // or:
 // "{unixTimestampMs}:{signatureAsHexString}:{clientChainId}" — signed over keccak256(timestamp || chainId)
-export const decodeSedaFastProof = (proof: string, chainId: string) => {
+export const decodeSedaFastProof = (proof: string, chainId?: string) => {
 	return Effect.gen(function* () {
 		try {
 			const decoded = Buffer.from(proof, "base64");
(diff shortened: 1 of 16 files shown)


workspace/data-proxy-sdk/src/services/decode-seda-fast-proof.ts
```diff
@@ -6,30 +6,53 @@ import {
...
-				Buffer.concat([unixTimestampBuffer, chainIdBytes]),
-			);
+			if (hasConfiguredChainId) {
+				if (!hasProofChainId) {
+					return yield* Effect.fail(
+						new FailedToDecodeSedaFastProofError({
+							error: "Proof missing chain id",
+						}),
+					);
+				}
```
Thomasvdam: Maybe I'm overlooking something, but I think we can allow the Data Proxy to have a chain ID configured while supporting Fast proofs without one.

That should make the update a little easier for external (not us) operators as simply updating the binary is enough.

workspace/data-proxy/src/cli/register.ts
```diff
@@ -59,7 +58,21 @@ export const registerCmd = new Command("register")
...
+				yield* Effect.logError(
+					"Network is required for registration. Pass --network or set network in the private key file.",
+				);
+				process.exit(1);
+			}
+
 			const network = defaultConfig[networkEnv];
+			if (network.chainId === undefined) {
+				yield* Effect.logError(
+					"Chain ID is required for registration. Pass --network or set network in the private key file.",
```
Thomasvdam: We only reach this error when the user did configure a network but it's not part of the default config, right? If that's true the error message should probably reflect that.

workspace/data-proxy/src/cli/run.ts
```diff
@@ -261,6 +267,14 @@ const configure = (
 				process.exit(1);
 			}
 
+			if (networkEnv === null) {
+				yield* Effect.logError(
+					"Network is required for registration check. Pass --network, set network in the private key file, or use --skip-registration-check.",
+				);
+				process.exit(1);
+			}
+
```
Thomasvdam: Maybe I'm reading it incorrectly but I think the lines above this change already run the registration check and I suppose that would have failed if there's no valid networkEnv? 

Regardless, `fastOnly` should probably also imply `--skip-registration-check`

workspace/data-proxy-sdk/src/services/decode-seda-fast-proof.ts
hacheigriega: I thought we may want the option to support testnet only or mainnet only even on Fast. 

But I guess what you're suggesting makes more sense given that Fast is removing chain ID from its proofs and that it would make upgrade easier.

workspace/data-proxy/src/cli/run.ts
hacheigriega: Correct, good catch. thanks

workspace/data-proxy/src/cli/register.ts
hacheigriega: Good point, done

workspace/data-proxy-sdk/src/services/decode-seda-fast-proof.ts
hacheigriega: We now verify proof's chain ID only if it is present and the data proxy has a chain ID configured.

Thomasvdam (approved):
We can do a minor version bump too :)

PR #147, opened 2026-07-07
feat(multi): let a sources query param prune the fan-out

jaspersagent:
A multi route runs every configured sub-fetch on every request, so a feed that only wants some venues still pays for the ones it discards. The worst case is a binance fetch for a symbol its stream does not carry, e.g. an equity perp like NVDAUSDT asked of the spot stream: no frame ever arrives, the fetch blocks the shared 3s price-wait timeout, and since the fan-out is concurrent the whole multi response is pinned at ~3s on every request (the cache never fills for a symbol that never streams, so it is not just a cold-start cost). This is exactly the seda-signal case where the program's `exchanges` input selects venues but the proxy fetched all of them anyway.

This adds a `sources` query param (comma-separated sub-fetch names) that restricts the fan-out to the named fetches. Skipped venues are never invoked and never subscribed, so mismatched symbols stop costing the timeout and stop accumulating dead WS subscriptions. An unknown or empty selection fails at the proxy with a 400 listing the configured source names, so a typo surfaces here rather than downstream as a confusing missing-source error. Requests without the param behave exactly as before, and the param rides in the signed request URL, so response signing and proof verification are unchanged.

Measured on the four-venue route (`/multi/:symbol/:market/:hydroTicker`): `multi/NVDA/110/xyz:NVDA` takes 3.01s without the param and 0.003s with `?sources=binance-futures,lighter,hydromancer`; BTC with no param still runs all four venues unchanged.

- No config or schema changes: valid source values are the route's existing `fetches[].name` list, validated per request.
- seda-signal follow-up (separate repo) appends `?sources=` from its `exchanges` input; the program change is backwards compatible against proxies without this support (param is simply ignored).

### DIFF
--- a/workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
+++ b/workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
@@ -9,6 +9,12 @@ import { replaceParams } from "../../utils/replace-params";
 // keyed by fetch name. Sub-fetch failures are non-fatal: the failing entry
 // carries an error and the rest still resolve. The combined object is returned
 // as one response for the proxy to sign.
+//
+// A `sources` query param (comma-separated fetch names) restricts the fan-out
+// to the named sub-fetches, so a caller that only wants some venues does not
+// pay for the rest (an unlisted symbol otherwise blocks on the price-wait
+// timeout). The param is part of the signed request URL, so selection needs no
+// signing changes. Without the param every configured fetch runs.
 export const handleMultiRequest = (
 	route: MultiModuleRoute,
 	params: Record<string, string>,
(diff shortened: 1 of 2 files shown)


workspace/data-proxy/src/controllers/proxy/handle-multi-request.ts
```diff
@@ -9,15 +9,54 @@ import { replaceParams } from "../../utils/replace-params";
...
+		const sourcesParam = new URL(request.url).searchParams.get("sources");
+		let selectedFetches = route.fetches;
+
+		if (sourcesParam !== null) {
+			const requested = sourcesParam
+				.split(",")
+				.map((name) => name.trim())
+				.filter((name) => name.length > 0);
+
+			const known = new Set(route.fetches.map((fetch) => fetch.name));
```
Thomasvdam: Using a `Set` is actually less performant in this scenario. Both arrays are small enough that the lookup 'gain' from a Set will never outweigh the construction cost.

jaspersagent: Fair point, at a handful of entries per route the Set never pays for itself. Switched both lookups to plain includes() over the arrays in 7066fe3; the known-names array also feeds the 400 error listing directly now.

Thomasvdam (approved):

PR #150, opened 2026-07-10
Volmex module

hacheigriega:
## Explanation of Changes

Volmex's WS client generally follows the structure of LO:TECH's, with a few differences.

I could not find a way to subscribe to specific tickers. As a result, the WebSocket client receives messages for all tickers, and it is the module's responsibility to keep track of "subscribed" tickers and update its cache only for those tickers. 

Volmex uses Engine.io as the persistent transport/session layer (handshake, heartbeats, framing), and Socket.io as the event layer on top for subscribing and receiving index price updates. Relevant docs:

- Volmex WebSocket doc https://docs.volmex.finance/index-data-api
- Engine IO doc https://socket.io/docs/v4/engine-io-protocol
- Socket IO doc https://socket.io/docs/v4/socket-io-protocol
- Short section on constructing WS frame https://socket.io/docs/v4/#what-is-the-overhead-of-the-socketio-protocol

This results in the following WS frames:

- `40`: Engine.IO message (4) + Socket.IO CONNECT (0) → join default namespace
- `42["event", payload]`: Engine.IO message (4) + Socket.IO EVENT (2) → emit an event

## Testing

Example config:
```json
{
	"modules": [
		{
			"type": "volmex",
			"name": "volmex",
			"baseUrl": "wss://ws-8jh89.volmex.finance",
			"volmexApiKeyEnvKey": "VOLMEX_API_KEY"
		}
	],
	"routes": [
		{
			"type": "volmex",
			"moduleName": "volmex",
			"path": "/:priceSymbol",
			"method": "GET",
			"fetchFromModule": "{:priceSymbol}"
		}
	]
}
```

API key can be obtained by logging in. Volmex stated that our credentials are configured not to expire, so it seems that every token we generate is permanent? This might be something we want to sort out before going into production. To obtain a token, set `VOLMEX_PASSWORD` and run:
```
JWT=$(curl -s --request POST \
  --url 'https://rest-v1.volmex.finance/auth/authorize' \
  --data "login=sedaprotocol&password=$VOLMEX_PASSWORD" \
  | jq -r '.d.access_token // .token // .accessToken // .jwt // empty')

echo $JWT
```

Run the program and make a query as follows:
```bash
bun start run --disable-proof -c volmex.jsonc

curl -s http://127.0.0.1:5384/proxy/EVIV7D | jq
```


Thomasvdam:
The expiration timestamp from the token is `86401783843097`, which gives us around 2000 years before we need to rotate. Seems fine.

Although when I decode the JWT in https://www.jwt.io/ it says the expiration date is invalid... As long as their BE accepts it I suppose 🤷

workspace/data-proxy/src/config/volmex-module-config.ts
```diff
@@ -0,0 +1,60 @@
...
+export type VolmexModulePriceFeed = v.InferOutput<
+	typeof VolmexModulePriceFeedSchema
+>;
+
+export const VolmexModuleConfigSchema = v.strictObject({
+	name: v.string(),
+	baseUrl: v.optional(v.string(), "wss://ws-8jh89.volmex.finance"),
+	priceFeeds: v.optional(v.array(VolmexModulePriceFeedSchema), []),
+	maxSymbolsPerRequest: v.optional(v.number(), 100),
+	volmexApiKey: v.string(),
```
Thomasvdam: This should probably be `volmexApiKeyEnv` or something similar like what we're using in other moduesl

Thomasvdam (commented):
Does the idle cleanup make sense?

As a concept I get that initially we introduced it on the Pyth module to prevent infrequent/one-off requests from consuming excessive resources. Especially in a world where there's a 'single' proxy servicing multiple clients. However, now that we're moving to more and more 'specialised' proxies that are deployed for a specific usecase I wonder how valuable that the cleanup still is...

Anyway, back to Volmex, I think here it's possibly even less useful. As I understand it we have to receive all messages anyway, so the only thing we're saving is a small pocket of memory in the cache. Maybe having a simpler module is worth this memory cost?

workspace/data-proxy/src/modules/volmex/ws-client.ts
```diff
@@ -0,0 +1,185 @@
...
+				}
+
+				const subscribeToIndices = (): void => {
+					socket.send(`42["${FETCH_INDICES_EVENT}",{}]`);
+				};
+
+				socket.on("open", () => {
+					Runtime.runSync(
+						runtime,
+						Effect.logInfo("Volmex WebSocket connected", { url }),
```
mariocao: We log the full `url`  but it contains `jwtToken=...`

This can leak API key in logs. Maybe log only base URL without query, or redact jwt param?

workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -42,6 +42,22 @@ interface PriceFeedWithSymbol extends ParsedFeedPayload {
...
+		return undefined;
+	}
+	const us =
+		typeof timestampUs === "string" ? Number(timestampUs) : timestampUs;
+	if (!Number.isFinite(us)) {
+		return undefined;
+	}
+	return nowMs - us / 1000;
+};
+
```
mariocao: Where is this used?

mariocao (dismissed):
I had 2 minor code comments and I also wanted to ask if it makes sense to add a handleRequest test as other modules.

Also, not super important, but no mention in readme.md. Does it make sense?

workspace/data-proxy/src/modules/volmex/ws-client.ts
hacheigriega: Good catch.. changed the key field in the module config to Effect `Redacted`.

workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
hacheigriega: Oops accidentally brought from a different branch

hacheigriega:
> Anyway, back to Volmex, I think here it's possibly even less useful. As I understand it we have to receive all messages anyway, so the only thing we're saving is a small pocket of memory in the cache. Maybe having a simpler module is worth this memory cost?

I checked that we get up to 50 messages per second (where each message only contains symbol, price, and timestamp), and when I leave the WebSocket on for 5 minutes I get ~300 unique symbols. So you're right, the memory usage is insignificant. Removed the idle cleanup structure and related structures.

hacheigriega:
> I had 2 minor code comments and I also wanted to ask if it makes sense to add a handleRequest test as other modules.

> Also, not super important, but no mention in readme.md. Does it make sense?

Certainly. Will also add READMEs to other modules incrementally.

workspace/data-proxy/src/modules/volmex/volmex.ts
```diff
@@ -0,0 +1,110 @@
...
+			yield* makeVolmexWebSocketService({
+				config,
+				runtime,
+				onPrice: updatePrice,
+			});
+
+			const start = () =>
+				Effect.gen(function* () {
+					yield* Effect.logInfo("Starting Volmex module");
+				}).pipe(Effect.annotateLogs("_name", "volmex"));
```
Thomasvdam: If the pattern we're using is `.start()` to start the module shouldn't this be responsible for creating the WS service? Now the module always starts immediately and the `start` method is just for fun :p

workspace/data-proxy/src/modules/volmex/ws-client.ts
```diff
@@ -0,0 +1,185 @@
...
+			new Promise((resolve) => {
+				const url = buildVolmexWebSocketUrl(
+					config.baseUrl,
+					config.volmexApiKey,
+				);
+				let namespaceConnected = false;
+
+				let socket: WebSocket;
+				try {
+					socket = new WebSocket(url);
```
Thomasvdam: Is there a reason we didn't use the Socket.io client?

Thomasvdam (commented):
Left a double comment, sorry

workspace/data-proxy/src/modules/volmex/volmex.ts
hacheigriega: Now it matches the Binance module style with the idempotent start https://github.com/sedaprotocol/seda-data-proxy/pull/150/commits/d7d374492fd8a6ae88c85d40a3b97e29f5b59e62

workspace/data-proxy/src/modules/volmex/ws-client.ts
hacheigriega: Was just following other module clients.. Switched to Socket.io client, and it cleaned up the code https://github.com/sedaprotocol/seda-data-proxy/pull/150/commits/d7d374492fd8a6ae88c85d40a3b97e29f5b59e62

workspace/data-proxy/src/modules/volmex/ws-client.ts
```diff
@@ -51,135 +16,100 @@ export type VolmexWebSocketServiceDeps = {
...
 
-		const reconnectSchedule = Schedule.spaced(
-			Duration.millis(config.reconnectDelayMs ?? 1000),
-		);
+			// Return a cleanup Effect that will be executed if the fiber
+			// running this effect is interrupted.
+			return Effect.sync(() => {
+				socket.removeAllListeners();
+				socket.close();
+			});
```
Thomasvdam: Not something to change as it's probably just syntax sugar, but this pattern feels like the `acquireRelease` pattern from Effect.

Thomasvdam (approved):

workspace/data-proxy/src/modules/volmex/ws-client.ts
hacheigriega: Addressed in https://github.com/sedaprotocol/seda-data-proxy/pull/150/commits/13198df3e4ad4d4f970e3ba6f94cd8d23153c906 for consistency with Binance pattern

PR #151, opened 2026-07-13
Addressing Pyth Lazer "client too slow" error

hacheigriega:
## Explanation of Changes

Some attempts at the "client too slow" error we get in Pyth Lazer module:

- Remove binary encoding format setting since we do not deal with binary encoded messages.
- Measure time lag between message timestamp and time at handling the message to get information about end-to-end lag.

### DIFF
--- a/workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
+++ b/workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
@@ -9,6 +9,8 @@ import {
 	Effect,
 	Either,
 	Layer,
+	Metric,
+	MetricBoundaries,
 	MutableHashMap,
 	Option,
 	Queue,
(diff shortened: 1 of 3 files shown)


workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -158,11 +176,26 @@ export const PythLazerModuleService = (config: PythLazerModuleConfig) =>
 
 			const handleStreamUpdatedMessage = (message: ParsedPayload) =>
 				Effect.gen(function* () {
-					yield* Effect.logTrace(
+					const nowMs = yield* Clock.currentTimeMillis;
+					const lagMs = lagMsFromTimestampUs(nowMs, message.timestampUs);
+
+					yield* Effect.logDebug(
```
Thomasvdam: Rather than logs I'd probably see if there's a metric that makes sense for this usecase: https://effect.website/docs/observability/metrics/

The time it takes to update process the message might also be interesting? It's probably tiny, but maybe it increases ever so slightly over time. Those kinds of developments are easier to track with a metric than a log with an (at this point) arbitrary threshold.

hacheigriega: Ah nice, changed to metrics. 
And yes, msg handling is probably very quick since it's just cache updates. I did add a duration tracker.

workspace/data-proxy/src/modules/pyth-lazer/pyth-lazer.ts
```diff
@@ -152,16 +189,18 @@ export const PythLazerModuleService = (config: PythLazerModuleConfig) =>
...
 			});
 
 			const handleStreamUpdatedMessage = (message: ParsedPayload) =>
 				Effect.gen(function* () {
-					yield* Effect.logTrace(
-						"Received stream updated message from Pyth Lazer client",
-						message,
-					);
+					const nowMs = yield* Clock.currentTimeMillis;
+					const lagMs = lagMsFromTimestampUs(nowMs, message.timestampUs);
```
Thomasvdam, line 199: Actually do we need to account for the fact that the timestamp is from US and the `nowMs` is UTC (I think).

hacheigriega: Pyth Lazer code doesn't clearly document this, but `timestampUs` seems to mean Unix timestamp in microseconds.

Thomasvdam (approved):

PR #153, opened 2026-07-14
PM Insights refactor

hacheigriega:
Small refactors before we deploy it

- Confine upstream fetch to the `GET /issuer/{symbol}` endpoint. The module only takes a single symbol as a request parameter per request.
- Added README and request handler tests

Link to PM Insights API docs: https://api.pminsights.com/docs

### DIFF
--- a/workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
+++ b/workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
@@ -1,0 +1,259 @@
+import { afterEach, describe, expect, it, mock } from "bun:test";
+import { Effect, Either, LogLevel, Logger } from "effect";
+import type { Route } from "../../config/config-parser";
+import type { PmInsightsModuleConfig } from "../../config/pm-insights-module-config";
+import { ModuleService } from "../module";
+import { PmInsightsModuleService, parseIssuerPrice } from "./pm-insights";
+
+const sampleIssuerResponse = {
+	updated_at: "2026-07-13 11:18:46",
+	info: {
+		id: "AVAN08",
+		display_name: "Anthropic",
+	},
+	price: {
+		price: 721.35,
+		is_derived: false,
+		currency: "USD",
+		ytd: 454.31,
+		ytd_roi: 170.13,
+	},
+	volume: {
+		three_months: {
+			bid: 8119675000.0,
+			ask: 20382777500.0,
+		},
+	},
+	valuation: null,
+};
+
+const baseConfig: PmInsightsModuleConfig = {
+	name: "pm-insights",
+	type: "pm-insights",
+	baseUrl: "https://api.pminsights.test/",
+	emailEnvKey: "PM_INSIGHTS_EMAIL",
+	passwordEnvKey: "PM_INSIGHTS_PASSWORD",
+	email: "user@test.com",
+	password: "secret",
+	tokenRefreshIntervalMinutes: 50,
+	tokenRetryIntervalMinutes: 5,
+};
+
+const routeFor = (fetchFromModule: string): Route =>
+	({
+		type: "pm-insights",
+		moduleName: "pm-insights",
+		fetchFromModule,
+		path: "/:symbol",
+		method: ["GET"],
+	}) as unknown as Route;
+
+const quiet = <A, E>(effect: Effect.Effect<A, E>) =>
+	effect.pipe(Logger.withMinimumLogLevel(LogLevel.None));
+
+const callHandle = (
+	config: PmInsightsModuleConfig,
+	params: Record<string, string>,
+	fetchFromModule = "{:symbol}",
+) => {
+	const program = Effect.gen(function* () {
+		const svc = yield* ModuleService;
+		return yield* svc.handleRequest(
+			routeFor(fetchFromModule),
+			params,
+			new Request("http://proxy.local/AVAN08"),
+		);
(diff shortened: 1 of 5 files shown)


workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -183,24 +229,20 @@ export const PmInsightsModuleService = (config: PmInsightsModuleConfig) =>
...
+								status: 400,
+							}),
+						);
 					}
 
-					const method = normalizeHttpMethod(request.method);
+					const upstreamUrl = new URL(
+						`issuer/${encodeURIComponent(symbol)}`,
+						config.baseUrl,
+					).toString();
```
Thomasvdam: Do we really want to constrain the proxy to just the `issuer/*` paths?

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
```diff
@@ -34,20 +60,49 @@ function parseTokenFromLoginBody(text: string): string | undefined {
...
-		case "DELETE":
-		case "PATCH":
-		case "HEAD":
-		case "OPTIONS":
-			return u;
-		default:
-			return "GET";
+/**
+ * Extracts the numeric issuer price from a PM Insights `/issuer/{symbol}` response body.
+ */
```
Thomasvdam: Ah maybe this is why we're limiting the endpoint? I personally think these kinds of checks are overstepping the function of the 'proxy', but that might just be a me thing.

Thomasvdam (commented):
Am I just not looking or right or did we remove the handling of a 4xx errors to trigger a reauthentication.

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
hacheigriega: It seems our API key is unauthorized to access some of the endpoints listed in the doc like:
```bash
$ curl -H "Authorization: Bearer $TOKEN" https://api.pminsights.com/feed/valuation/AVAN08

{"detail":"Not authorized for reference data, please contact support@pminsights.com."}%
```
And I thought our focus was on prices of specific assets (issuers here) instead of sector / investor / overall market data. I don't have a strong opinion on this, though.

Thomasvdam: I guess it comes down to a philosophy thing. If we're just building this for us and our usecase I think it's fine to limit it. If we want it to be more generally applicable we should probably let it behave more like a proxy.

hacheigriega:
> Am I just not looking or right or did we remove the handling of a 4xx errors to trigger a reauthentication.

You are never wrong :o Added it back. I forgot how to simulate that scenario, but I added a test case.

workspace/data-proxy/src/modules/pm-insights/pm-insights.ts
hacheigriega: I see what you mean by this "overstepping the function of the proxy." Changed to support any upstream paths. Wouldn't hurt to support any paths anyways.

workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
```diff
@@ -0,0 +1,263 @@
...
+};
+
+const urlOf = (input: URL | RequestInfo): string =>
+	input instanceof URL
+		? input.toString()
+		: input instanceof Request
+			? input.url
+			: String(input);
+
+const mockPmInsightsFetch = (handlers?: {
```
Thomasvdam, line 81: If I'm not mistaken we can use dependency injection to mock the HTTP client instead of relying on mutating the global `fetch`.

Not something urgent, but probably good to at least leave a TODO

workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+		const response = await callHandle(
+			baseConfig,
+			{},
+			"sectors",
+			"http://proxy.local/sectors",
+		);
+		expect(response.status).toBe(200);
+		expect(await response.json()).toEqual([{ id: "tech" }]);
+	});
```
Thomasvdam: This is (in my opinion) a classic LLM test. It tests that some behaviour that was present in the history of the PR is no longer present, but if were to look at the absolute set of changes that this PR introduces this test makes very little sense :D

Thomasvdam (approved):

mariocao (approved):

workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
hacheigriega: Added TODO. Will make sure to address it soon

workspace/data-proxy/src/modules/pm-insights/pm-insights.test.ts
hacheigriega: Good catch.. So if I understood correctly, this sort of things happens because LLM tends to look at history without recognizing when it is no longer relevant?

Thomasvdam: It could be history, or simply when we prompt "no longer do thing X" it doesn't 'know' that the software never had to do X anyway and creates a test for it.

PR #155, opened 2026-07-21
Pyth Lazer module refactor

hacheigriega:
- Support multiple channels per module instance through config refactoring.
  - A single module can now support multiple routes with different channels.
- Configures number of connections for Pyth Client at 3 to match the three URLs we use. Pyth SDK had been using its default value of 4.
- Adds module README
- Redact API key from error message from the SDK that we log

### DIFF
--- a/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
+++ b/workspace/data-proxy/src/config/pyth-lazer-module-config.ts
@@ -1,6 +1,18 @@
+import type { Channel } from "@pythnetwork/pyth-lazer-sdk";
 import { Duration, Effect, Option } from "effect";
 import * as v from "valibot";
 import { RouteSchema } from "./route-config";
+
+export const PYTH_LAZER_DEFAULT_CHANNEL = "fixed_rate@200ms" as const;
+
+const PYTH_LAZER_CHANNELS = [
+	"real_time",
+	"fixed_rate@50ms",
+	"fixed_rate@200ms",
+	"fixed_rate@1000ms",
+] as const satisfies readonly Channel[];
+
+export const PythLazerChannelSchema = v.picklist(PYTH_LAZER_CHANNELS);
 
 export const PythLazerModuleConfigSchema = v.strictObject({
 	name: v.string(),
@@ -8,16 +20,8 @@ export const PythLazerModuleConfigSchema = v.strictObject({
 		v.object({
 			name: v.string(),
 			id: v.number(),
+			channel: v.optional(PythLazerChannelSchema, PYTH_LAZER_DEFAULT_CHANNEL),
 		}),
-	),
-	channel: v.optional(
-		v.picklist([
-			"real_time",
-			"fixed_rate@50ms",
-			"fixed_rate@1000ms",
-			"fixed_rate@200ms",
-		]),
-		"fixed_rate@200ms",
 	),
 	maxFeedsPerRequest: v.optional(v.number(), 100),
 	pythLazerApiKeyEnvKey: v.string(),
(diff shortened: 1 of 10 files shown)


workspace/data-proxy/src/controllers/proxy/handle-multi-request.test.ts
```diff
@@ -15,11 +15,16 @@ const handlers = (
...
+// variant (e.g. upstream, hydromancer), so narrow before reading it.
 const echoHandlers = handlers((route, _params, _request, body) =>
 	Effect.succeed(
 		new Response(
-			JSON.stringify({ fetchFromModule: route.fetchFromModule, body }),
+			JSON.stringify({
+				fetchFromModule:
+					"fetchFromModule" in route ? route.fetchFromModule : undefined,
+				body,
+			}),
```
Thomasvdam, line 27: I don't get why we need this? If `route` doesn't have a key `fetchFromModule` it will be `undefined` anyway 🤔

workspace/data-proxy/src/config/pyth-lazer-module-config.ts
```diff
@@ -1,24 +1,28 @@
...
+
+export const PythLazerChannelSchema = v.picklist(PYTH_LAZER_CHANNELS);
+
 export const PythLazerModuleConfigSchema = v.strictObject({
 	name: v.string(),
 	priceFeedIds: v.array(
 		v.object({
 			name: v.string(),
 			id: v.number(),
+			channel: v.optional(PythLazerChannelSchema),
```
Thomasvdam: It's a little weird that we're doing optional with a default for the route, but here we leave the rest of the app to handle a default?

workspace/data-proxy/src/config/pyth-lazer-module-config.ts
```diff
@@ -1,24 +1,28 @@
...
 	),
-	channel: v.optional(
-		v.picklist([
-			"real_time",
-			"fixed_rate@50ms",
-			"fixed_rate@1000ms",
-			"fixed_rate@200ms",
-		]),
-		"fixed_rate@200ms",
-	),
```
mariocao, line 21: This is make old configs not compatible. Just flagging it because of `strictObject`. :)

mariocao (approved):
It looks good to me.
if it is not backwards compatible, it should be a major upgrade, right?

workspace/data-proxy/src/config/pyth-lazer-module-config.ts
hacheigriega: Thank you. Let me know if it is better to make it NOT strict

Thomasvdam: No Mario raised a good point that it's better to error hard instead of silently applying a default

workspace/data-proxy/src/controllers/proxy/handle-multi-request.test.ts
hacheigriega: I was confused by this too, but this is essentially just to get the response in the shape the tests expect. It surfaces fetchFromModule to the top level because it is the only thing that the tests check.

hacheigriega:
> It looks good to me. if it is not backwards compatible, it should be a major upgrade, right?

So v3.0.0, right?

workspace/data-proxy/src/controllers/proxy/handle-multi-request.test.ts
Thomasvdam: Sorry, I meant the benefit of `fetchFromModule: "fetchFromModule" in route ? route.fetchFromModule : undefined,` over `fetchFromModule: route.fetchFromModule`

hacheigriega: I think that does not work because `fetchFromModule` is not on every Route variant?

PR #156, opened 2026-07-23
Expand comma-separated coins in Hydro request body

hacheigriega:
## Motivation

To support module route like the following
```json
{
  "type": "multi",
  "path": "/multi/:hydroTickers",
  "method": ["GET"],
  "fetches": [
    {
      "name": "hydromancer",
      "moduleName": "hydro",
      "type": "hydromancer",
      "body": "{\"type\":\"assetContext\",\"coins\":\"{:hydroTickers}\"}"
    }
  ]
}
```

we need `hydroTickers` given as a path parameter to expand into multiple tickers.

## Explanation of Changes

Expand comma-delimited `coins` values in Hydromancer assetContext bodies into individual tickers.

## Testing

Added some test cases.

### DIFF
--- a/workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
+++ b/workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
@@ -87,14 +87,23 @@ export const HydromancerModuleService = (config: HydromancerModuleConfig) =>
 						return yield* executeHydromancerRestRequest(config, body);
 					}
 
-					// We need to know if the request is for a single coin or a batch of coins as the response shape is different.
+					// Normalize coins (string or string[]) and expand comma-separated
+					// values (e.g. "BTC,ETH" or ["BTC,ETH"] from multi-route path params).
+					const expandCoins = (tokens: string | string[]) =>
+						(typeof tokens === "string" ? [tokens] : tokens).flatMap((token) =>
+							token
+								.split(",")
+								.map((coin) => coin.trim())
+								.filter((coin) => coin.length > 0),
+						);
+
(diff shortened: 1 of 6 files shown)


gluax (approved):
LGTM but would wait for someone more familiar to review ^^

workspace/data-proxy/src/modules/hydromancer/hydromancer.ts
```diff
@@ -87,14 +87,23 @@ export const HydromancerModuleService = (config: HydromancerModuleConfig) =>
...
-					// We need to know if the request is for a single coin or a batch of coins as the response shape is different.
+					// Expand comma-separated path params (e.g. coins:["BTC,ETH"] from
+					// multi routes) into individual tickers.
+					const expandCoins = (tokens: string[]) =>
+						tokens.flatMap((token) =>
+							token
+								.split(",")
+								.map((coin) => coin.trim())
+								.filter((coin) => coin.length > 0),
+						);
```
Thomasvdam: Please bear with me while I rant a little. You can skip to the last paragraph if you're not interested in me complaining :D 

I think this perfectly highlights how poor the multi module is (and by extension our module architecture). We have to make adaptations in the source modules to accommodate the multi module. So the multi module is source aware, and now the sources are (somewhat) multi aware. :(

Anway, from a practical standpoint I think it's more user friendly if we allow `coins` to be either an array of strings (current implementation), or a single string with comma delimited strings. My main reason is that it's easy to forget in the config that you have to wrap the parameter in `[...]`.

Thomasvdam (approved):

PR #157, opened 2026-07-26
Fix: multi skipped symbols

gluax:
## Motivation

So the issue is:

If we didn't need any assets from a source, the request would fail. This could happen in two scenarios:
1. It was the only asset in a group that did not have a source
2. All the assets in a group missed a source

Unlikely, but as seen on staging, it could happen. So IMO better to have then to cause a blip/production issue.

## Explanation of Changes

Allows `_` to be sent in place of a skipped asset. As a consequence, this works on `csv` style as well, but the OP only sends a singular `/_/` if the entire group has no assets for that source.

## Testing

New tests added.

## Related PRs and Issues

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,8 @@ A **multi** route fans out to several module handlers in one request. Each entry
 - **`body`** — forwarded to modules that expect a request body (for example Hydromancer `assetContext`).
 - Sub-fetch failures are **non-fatal**: the failing source appears as an `{ "error": "...", "status": ... }` entry; other sources still resolve. The multi route itself returns HTTP `200`.
 - Fetch `name` values must be **unique** within a route (validated at startup).
+- A path element of **`_`** means "no asset for this source" and is dropped before templates are filled, so `/multi/BTC,_,SOL/1,2,3/BTC,ETH,SOL` asks the first source for BTC and SOL only. Only whole elements count, so `BTC_USD` is unaffected.
+- A fetch whose param is left **entirely empty** is skipped without calling its module; its entry is `[]` for the price modules, `{}` for Hydromancer, `null` for `pm-insights` and `chainlink-streams`. `_` and `sources` compose: `sources` selects, `_` subtracts.
 
 Example — Binance spot, Lighter perp ticker, and Hydromancer asset context in one call:
 
@@ -439,6 +441,9 @@ Test locally (disable proof verification for easier debugging):
 
 ```bash
 curl -s "http://127.0.0.1:5384/proxy/multi/BTC/1/BTC" | jq .
+
+# Skip Binance for this request
+curl -s "http://127.0.0.1:5384/proxy/multi/_/1/BTC" | jq .
 ```
 
 Example response shape:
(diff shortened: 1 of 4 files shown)


mariocao (approved):
Approving, but we should definitely refactor the module.
First thing I would do is to use POST and its body instead of query params.
@Thomasvdam most probably has other good ideas. :)

Thomasvdam (approved):
Like Mario said I think for now this is ok, but it's also a strong signal we should refactor the multi module.

@hacheigriega I think this would be a great task to work on this week. I don't want to tell you how to do (that's the fun part :P), but of course I will share my 20 cents because I'm a 10x opinion-haver.

At the root there's a problem with this module in that it needs to be aware of how other modules work. There's a parallel here with how the proxy in general is becoming more aware of how different backends work, but lets leave that for now. So now every time we add a module we'd also have to update the multi module and we have to update the config of the deployed multi module as well. Ideally this functionality of batching calls 'just works' when you enable it without additional maintenance.

Conceptually the multi module is very similar to [GraphQL](https://graphql.org/), which allows the client to do 1 request to the backend that behind the scenes retrieves data from different sources. I think GraphQL here is overkill, but we can take inspiration from the concept.

Some ideas that may or may not be useful:
- Like Mario said, POST probably makes the most sense. The payload can then specify what to fetch from where.
- Maybe it shouldn't be a module but instead a core feature on a dedicated endpoint, similar to the health/info under `/status/*`.
- We could have it automatically use the route config of the proxy to 'know' what multi routes are available instead of a separate config.
- Regardless of whether it is a module or core feature, we should probably allow running it in parallel with the current multi module to make the transition easier.

gluax:
I super agree and would prefer a POST request here as well. This was me trying to bandaid the solution for now so as to roll this out quickly for moon :smiling_face_with_tear:

README.md
```diff
@@ -388,6 +388,8 @@ A **multi** route fans out to several module handlers in one request. Each entry
 - **`body`** — forwarded to modules that expect a request body (for example Hydromancer `assetContext`).
 - Sub-fetch failures are **non-fatal**: the failing source appears as an `{ "error": "...", "status": ... }` entry; other sources still resolve. The multi route itself returns HTTP `200`.
 - Fetch `name` values must be **unique** within a route (validated at startup).
+- A path element of **`_`** means "no asset for this source" and is dropped before templates are filled, so `/multi/BTC,_,SOL/1,2,3/BTC,ETH,SOL` asks the first source for BTC and SOL only. Only whole elements count, so `BTC_USD` is unaffected.
```
hacheigriega, line 391: Just wondering what would be the use case for such a path? Why can't we just call `/multi/BTC,SOL/1,2,3/BTC,ETH,SOL`?

gluax: So as seen from staging. What if all the assets in a batch are missing binance? What if there is only a singular asset in the batch and it's missing binance?

Essentially those are the cases we ran into on staging. That broke since the Data Proxy before hand didn't accept those.

As such we needed to be able to do /_/ in those cases.

PR #161, opened 2026-07-31
Multi endpoint

hacheigriega:
## Motivation

Support multi routing (multiple sub-requests across different modules in a single request) with minimal additional configuration.

## Explanation of Changes

If multi endpoint is enabled in the config, users can start making requests like:
```bash
curl -sS -X POST 'http://127.0.0.1:5384/proxy/multi' \
  -H 'content-type: application/json' \
  -d '{
    "requests": [
      { "id": "binance-futures", "path": "/binance-futures/BTCUSDT,ETHUSDT" },
      { "id": "lighter", "path": "/lighter/1,0" },
      {
        "id": "hydromancer",
        "path": "/hydromancer/asset-context",
        "method": "POST",
        "body": { "type": "assetContext", "coins": "BTC,ETH" }
      }
    ]
  }'
```
`path` is matched against the proxy's configured routes, just like paths clients already call one-by-one.

We keep the existing multi route structure until the infra is upgraded. We obviously block calling a multi route through the multi endpoint.

### DIFF
--- a/workspace/data-proxy/src/config/config-parser.ts
+++ b/workspace/data-proxy/src/config/config-parser.ts
@@ -48,7 +48,7 @@ import { type Modules, ModulesSchema } from "./module-config";
 import {
 	DEFAULT_MULTI_CONCURRENCY,
 	DEFAULT_MULTI_ENDPOINT_PATH,
-	DEFAULT_MULTI_MAX_REQUESTS,
+	DEFAULT_MULTI_MAX_SUB_REQUESTS,
 	MultiEndpointSchema,
 } from "./multi-endpoint-config";
 import {
@@ -186,7 +186,7 @@ const ConfigSchema = v.strictObject(
 		multiEndpoint: v.optional(MultiEndpointSchema, {
 			enable: false,
 			path: DEFAULT_MULTI_ENDPOINT_PATH,
-			maxRequests: DEFAULT_MULTI_MAX_REQUESTS,
+			maxSubRequests: DEFAULT_MULTI_MAX_SUB_REQUESTS,
 			concurrency: DEFAULT_MULTI_CONCURRENCY,
 		}),
 	},
--- a/workspace/data-proxy/src/config/config-parser.ts
+++ b/workspace/data-proxy/src/config/config-parser.ts
@@ -11,6 +11,7 @@ import {
 	DEFAULT_VERIFICATION_MAX_RETRIES,
 	DEFAULT_VERIFICATION_RETRY_DELAY,
 } from "../constants";
+import { normalizePath } from "../utils/match-route-path";
 import { replaceParams } from "../utils/replace-params";
 import {
 	type BinanceModuleRoute,
@@ -44,6 +45,12 @@ import {
 	validateLoTechModuleRoute,
 } from "./lo-tech-module-config";
 import { type Modules, ModulesSchema } from "./module-config";
+import {
+	DEFAULT_MULTI_CONCURRENCY,
+	DEFAULT_MULTI_ENDPOINT_PATH,
+	DEFAULT_MULTI_MAX_REQUESTS,
+	MultiEndpointSchema,
+} from "./multi-endpoint-config";
 import {
 	type MultiModuleRoute,
 	MultiModuleRouteSchema,
@@ -176,6 +183,12 @@ const ConfigSchema = v.strictObject(
 				root: "status",
 			},
 		),
+		multiEndpoint: v.optional(MultiEndpointSchema, {
+			enable: false,
+			path: DEFAULT_MULTI_ENDPOINT_PATH,
+			maxRequests: DEFAULT_MULTI_MAX_REQUESTS,
+			concurrency: DEFAULT_MULTI_CONCURRENCY,
+		}),
 	},
 	UNKNOWN_ATTRIBUTE_ERROR,
 );
@@ -251,6 +264,21 @@ export const parseConfig = (
 				),
 				hasWarnings,
 			];
+		}
+
+		if (config.multiEndpoint.enable) {
+			const multiPath = normalizePath(config.multiEndpoint.path);
+
+			for (const route of config.routes) {
+				if (normalizePath(route.path) === multiPath) {
+					return [
+						Result.err(
+							`Route ${route.path} collides with multiEndpoint.path; rename the route or multiEndpoint.path`,
+						),
+						hasWarnings,
+					];
+				}
+			}
 		}
 
 		// Check if the environment variables required by the status endpoint are available.
--- a/workspace/data-proxy/src/config/multi-endpoint-config.ts
+++ b/workspace/data-proxy/src/config/multi-endpoint-config.ts
@@ -1,0 +1,59 @@
+import * as v from "valibot";
+
+export const DEFAULT_MULTI_ENDPOINT_PATH = "multi";
+export const DEFAULT_MULTI_MAX_REQUESTS = 20;
+export const DEFAULT_MULTI_CONCURRENCY = 5;
+
+export const MultiEndpointSchema = v.strictObject({
+	enable: v.optional(v.boolean(), false),
+	path: v.optional(v.string(), DEFAULT_MULTI_ENDPOINT_PATH),
+	maxRequests: v.optional(
+		v.pipe(
+			v.number(),
+			v.integer(),
+			v.minValue(1, "maxRequests must be a positive integer"),
+		),
+		DEFAULT_MULTI_MAX_REQUESTS,
+	),
+	concurrency: v.optional(
+		v.pipe(
+			v.number(),
+			v.integer(),
+			v.minValue(1, "concurrency must be a positive integer"),
+		),
+		DEFAULT_MULTI_CONCURRENCY,
+	),
+});
+
+export type MultiEndpoint = v.InferOutput<typeof MultiEndpointSchema>;
+
+// A single request inside a multi endpoint body. `path` is matched against the
+// proxy's configured routes (same public paths clients already call one-by-one).
+export const MultiEndpointRequestSchema = v.strictObject({
+	path: v.pipe(v.string(), v.minLength(1, "path must not be empty")),
+	method: v.optional(v.string(), "GET"),
+	query: v.optional(
+		v.record(v.string(), v.union([v.string(), v.array(v.string())])),
+	),
+	// String is forwarded as-is; objects/arrays are JSON-stringified.
+	body: v.optional(v.unknown()),
+});
+
+export type MultiEndpointRequest = v.InferOutput<
+	typeof MultiEndpointRequestSchema
+>;
+
+export const MultiEndpointRequestBodySchema = v.pipe(
+	v.record(
+		v.pipe(v.string(), v.minLength(1, "id must not be empty")),
+		MultiEndpointRequestSchema,
+	),
+	v.check(
+		(entries) => Object.keys(entries).length >= 1,
+		"request body must contain at least one entry",
+	),
+);
+
+export type MultiEndpointRequestBody = v.InferOutput<
+	typeof MultiEndpointRequestBodySchema
+>;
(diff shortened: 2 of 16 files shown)


workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
```diff
@@ -0,0 +1,457 @@
...
+
+		const seenIds = new Set<string>();
+		for (const entry of multiBody.requests) {
+			if (seenIds.has(entry.id)) {
+				return createBadRequestResponse(
+					`Duplicate request id "${entry.id}"; ids must be unique within a multi request`,
+				);
+			}
+			seenIds.add(entry.id);
+		}
```
Thomasvdam: I'm not sure if this is possible to hit? If I'm not mistaken a JSON serialised object with the same key multiple times will deserialise into an object with the last occurrence of the key in the string form.? But probably good to research quickly.

Thomasvdam (commented):
Sorry I don't have time for a thorough review right now, but I have 2 questions that might be worth pondering:

Can the multi path call itself? We should probably disallow that.
Maybe good to see if Elysia supports calling itself programmatically? That would save a bunch of routing code.

I'll take a better look next week and actually research my comments before making them :P

hacheigriega:
> Can the multi path call itself? We should probably disallow that. Maybe good to see if Elysia supports calling itself programmatically? That would save a bunch of routing code.

Thanks.. I wanted to open the PR yesterday, but it ended up taking longer :/
We disallow the now legacy multi route. And the multi endpoint cannot call itself because it is not a module nor a route.
 
Calling Elysia itself is an interesting idea. It would mean duplicate work for pre- and post- request handling like proof check and signing, though. I'm thinking I could create helper functions to address the duplicate code issue. Let me know what you think.

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
hacheigriega: I think so. It's not the key but the value that we are checking for duplication here, right?

Thomasvdam:
> dCalling Elysia itself is an interesting idea. It would mean duplicate work for pre- and post- request handling like proof check and signing, though. I'm thinking I could create helper functions to address the duplicate code issue. Let me know what you think.

true, I didn't consider that :) That makes me think that the current approach makes more sense.

workspace/data-proxy/src/config/config-parser.ts
```diff
@@ -253,6 +255,29 @@ export const parseConfig = (
...
+			const multiPath = normalizePath(config.multiEndpoint.path);
+
+			for (const route of config.routes) {
+				const routeMethods = getHttpMethods(route.method).map((m) =>
+					m.toUpperCase(),
+				);
+				if (
+					normalizePath(route.path) === multiPath &&
+					routeMethods.includes("POST")
+				) {
```
Thomasvdam: I think it would be kinda confusing to allow the same path with a different method to do something completely different. Probably clearer to simply disallow having the same path as the multi path.

workspace/data-proxy/src/proxy-server.ts
```diff
@@ -181,15 +185,73 @@ export const startProxyServer = (
 				}
 			});
 
+			// If enabled, multi endpoint fans out to configured module/upstream
+			// routes by path.
+			const multiEndpoint = config.multiEndpoint ?? DEFAULT_MULTI_ENDPOINT;
```
Thomasvdam: Why do we resolve the default here instead of in the config?

workspace/data-proxy/src/config/multi-endpoint-config.ts
```diff
@@ -0,0 +1,53 @@
...
+});
+
+export type MultiEndpointRequest = v.InferOutput<
+	typeof MultiEndpointRequestSchema
+>;
+
+export const MultiEndpointRequestBodySchema = v.strictObject({
+	requests: v.pipe(
+		v.array(MultiEndpointRequestSchema),
+		v.minLength(1, "requests must contain at least one entry"),
```
Thomasvdam: A bit 🥔 🥔 , but was there a reason to do

```ts
[
  { id: 1, ... },
  { id: 2, ...},
]
```

instead of 

```ts
{
  1: { ... },
  2: { ... },
}
```

workspace/data-proxy/src/proxy-server.ts
```diff
@@ -181,15 +185,73 @@ export const startProxyServer = (
...
+					{
+						config: {},
+						parse: ({ request }) => request.text(),
+					},
+				);
+
+				Runtime.runSync(
+					runtime,
+					Effect.logInfo(
+						`Multi endpoint: POST /${proxyGroup === "" || proxyGroup.endsWith("/") ? proxyGroup : `${proxyGroup}/`}${multiPath.replace(/^\//, "")}`,
```
Thomasvdam: Kinda funny that there's 2 solutions to the same problem, 1 with a regex and 1 with a check :P I would just do the regex solution on both strings.

workspace/data-proxy/src/controllers/proxy/execute-route.ts
```diff
@@ -0,0 +1,209 @@
...
+				Effect.gen(function* () {
+					yield* Effect.logDebug("Handling multi request");
+					return yield* handleMultiRequest(
+						multiModuleRoute,
+						params,
+						request,
+						moduleHandlers,
+					);
+				}),
+			),
```
Thomasvdam, line 64: Maybe we should also mark this as legacy?

workspace/data-proxy/src/controllers/proxy/execute-route.ts
```diff
@@ -0,0 +1,209 @@
...
+							}),
+						);
+					}
+
+					yield* Effect.logDebug(`Handling ${moduleRoute.type} request`);
+
+					const moduleBody =
+						moduleRoute.type === "hydromancer"
+							? Option.getOrElse(body, () => "")
+							: Option.getOrUndefined(body);
```
Thomasvdam: This is a code smell. Maybe not something to fix now, but we should note this and see if we can in the future make it so that router doesn't need to know about the destinations.

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
```diff
@@ -0,0 +1,324 @@
...
+
+		// Run each request concurrently.
+		const entries = yield* Effect.forEach(
+			multiBody.requests,
+			(entry) => runRequest(entry, inputParams),
+			{ concurrency: "unbounded" },
```
Thomasvdam: Maybe unbounded is too optimistic? Now every request can spawn `maxRequests` new requests. Maybe a separate config value for this, or start with a lower cap of ~5?

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
Thomasvdam: In my quick read I assumed the body was `{ idA: { paramsA }, idB: { paramsB } }`. So I guess this question is replaced by my question about the format :D

workspace/data-proxy/src/utils/match-route-path.ts
```diff
@@ -0,0 +1,46 @@
...
+	for (let i = 0; i < patternParts.length; i++) {
+		const patternPart = patternParts[i];
+		const pathPart = pathParts[i];
+
+		if (patternPart.startsWith(":")) {
+			const key = patternPart.slice(1);
+			if (key.length === 0) {
+				return null;
+			}
+			params[key] = decodeURIComponent(pathPart);
```
Thomasvdam: (AI review comment)

`decodeURIComponent` throws on invalid characters and since these are arbitrary strings in a POST body we should probably be a little careful?

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
```diff
@@ -0,0 +1,324 @@
...
+		return undefined;
+	}
+	if (typeof body === "string") {
+		return body;
+	}
+	return JSON.stringify(body);
+};
+
+const findMatchingRoute = (
+	eligibleRoutes: readonly Config["routes"][number][],
```
Thomasvdam: Doesn't `[number]` 'unpack' the array only for us to package it again with `[]` at the end?

workspace/data-proxy/src/utils/match-route-path.ts
Thomasvdam, line 1: I don't think Elysia exposes the router or helper methods, but maybe we can simply copy the source? Given that we should handle matching and params in the same way.

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
hacheigriega: true, good point

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
hacheigriega: Ah, correct, and I realized it has to be used with `Readonly<T>` instead of `readonly T`

workspace/data-proxy/src/controllers/proxy/execute-route.ts
hacheigriega: I removed it as I realized undefined and empty string behave the same way for Hydromancer module. Thanks.

Thomasvdam (approved):

hacheigriega:
Thanks again for the review @Thomasvdam 
I was getting confused over the Elysia (Memoirist to be precise) router code. I'm thinking to create a separate issue or PR to address that one.
I would really appreciate if you could take a gander at the last commit where I added support for wildcard for multi routes and prevent certain request headers from being forwarded?

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
```diff
@@ -46,6 +46,29 @@ const serializeRequestBody = (body: unknown): string | undefined => {
 	return JSON.stringify(body);
 };
 
+// Parent headers that must not leak onto sub-requests
+const NON_FORWARDED_HEADERS = new Set([
```
Thomasvdam: What about `content-type`?

Thomasvdam: Shouldn't we simply make the headers an optional param in the multi body and disregard any headers on the original request?

workspace/data-proxy/src/utils/match-route-path.ts
```diff
@@ -8,10 +9,39 @@ export const matchRoutePath = (
...
+			return null;
+		}
+
+		const params = matchSegments(prefixParts, pathParts);
+		if (params === null) {
+			return null;
+		}
+
+		// Capture the remainder of the path as the `*` value.
+		params["*"] = pathParts.slice(prefixParts.length).join("/");
```
Thomasvdam, line 30: This param is not decoded with `decodeURIComponent` unlike the rest. Does this matter?

workspace/data-proxy/src/controllers/proxy/handle-multi-endpoint-request.ts
hacheigriega: That sounds better to me. Made the headers an optional param in the multi body

workspace/data-proxy/src/utils/match-route-path.ts
hacheigriega: I see this is an important question. My reasoning is: We do not decode the wildcard value because it is a path that we want to pass to the upstream as is, unlike non-wildcard parameter values, which we want to extract and use ourselves.

hacheigriega: I checked that this is also what Elysia / Memoirist does

Thomasvdam:
Sorry I forgot to add that it was a good catch with the headers and the catch all param! I completely missed that 🦅 👀

